PR #11, opened 2016-11-20
Build instructions for GraphiQL example

andreas:
To run a sample GraphQL server also serving GraphiQL, do the following:

```bash
git checkout git@github.com/andreas/ocaml-graphql-server.git
cd ocaml-graphql-server
opam pin add graphql-server .
cd examples
ocamlbuild -use-ocamlfind server.native && ./server.native
```

Now open [http://localhost:8080](http://localhost:8080).

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,17 @@ Current feature set:
 
 ### GraphiQL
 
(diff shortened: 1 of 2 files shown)


sgrove:
This worked great for me - thanks!

PR #15, opened 2016-12-18
Lwt and Async support

### LINKED ISSUE #2: Asynchronous resolvers
It should be possible to return a `Lwt.t` or `Async.t` from the `resolve`-function of a field. A likely solution is to functorize the execution over an `IO` monad.

andreas:
This PR adds support for Lwt and Async in resolvers by adding two subpackages `graphql.lwt` and `graphql.async`, which contain the modules `Graphql_lwt.Schema` and `Graphql_async.Schema`, respectively. These modules are similar to `Graphql.Schema`, but has the function `io_field`:

```ocaml
val io_field : string ->
               typ:('ctx, 'a) typ ->
               args:('a io, 'b) Arg.arg_list ->
               resolve:('ctx -> 'src -> 'b) ->
               ('ctx, 'src) field
```

where `'a io` is `'a Lwt.t` (`Graphql_lwt`) or `'a Deferred.t` (`Graphql_async`). Note that there's still the function `field` for "direct" fields. Example:

```ocaml
let schema = Graphql_lwt.Schema.(schema
    ~fields:[
      field "direct_string"
        ~typ:(non_null string)
        ~args:Arg.[]
        ~resolve:(fun () () -> "foo")
      ;
      io_field "io_int"
        ~typ:(non_null int)
        ~args:Arg.[]
        ~resolve:(fun () () -> Lwt.return 42)
    ]
)
```

See `examples/server.ml` for a more complete example.

@kennetpostigo @sgrove: Any feedback is greatly appreciated 😄

~~Note that I will most likely reorganized the changes in this PR into multiple commits/PRs -- making a PR now was mostly to share these changes sooner rather than later.~~
Done!

Closes #2.


kennetpostigo:
Just read through `server.ml`, `async_test.ml`, and `lwt_test.ml` everything looks really solid! Great work @andreas !

PR #24, opened 2017-02-04
Mutation support

### LINKED ISSUE #4: Mutation support
See https://facebook.github.io/graphql/#sec-Executing-Operations.

andreas:
[GraphiQL demo of mutation support](https://ocaml-graphql.herokuapp.com/?query=mutation%20%7B%0A%20%20add_number(number%3A%2042)%0A%7D&operationName=undefined) ✨ 

This PR implements support for mutations. The type of `Graphql.schema` has been updated from

```ocaml
(* Previous type *)
val schema : fields:('ctx, unit) field list ->
             'ctx schema
```
to
```ocaml
(* New type *)
val schema : ?mutation_name:string ->
             ?mutations:('ctx, unit) field list ->
             ?query_name:string ->
             ('ctx, unit) field list ->
             'ctx schema
```

The optional argument `~mutations` will be exposed as fields on the mutation root of the schema. Top-level mutation fields are executed serially as [described in the spec](https://facebook.github.io/graphql/#sec-Executing-Operations).

#### Example schema

```ocaml
let numbers = ref []

let schema = Schema.(schema [
      field "numbers"
        ~typ:(non_null (list (non_null int)))
        ~args:Arg.[]
        ~resolve:(fun () () -> !numbers)
    ]
    ~mutations:[
      field "add_random"
        ~typ:(non_null (list (non_null int)))
        ~args:Arg.[arg "number" ~typ:(non_null int)]
        ~resolve:(fun () () number ->
          numbers := number::!numbers;
          !numbers
        )
    ]
)
```

closes #4

### DIFF
--- a/src/graphql_intf.ml
+++ b/src/graphql_intf.ml
@@ -12,8 +12,11 @@ module type Schema = sig
 
   (** {3 Constructors } *)
 
-  val schema : fields:('ctx, unit) field list ->
-              'ctx schema
+  val schema : ?mutation_name:string ->
+               ?mutations:('ctx, unit) field list ->
+               ?query_name:string ->
+               ('ctx, unit) field list ->
+               'ctx schema
 
   val obj : ?doc:string ->
             string ->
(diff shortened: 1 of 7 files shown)


sgrove:
At first glance this seems sensible - looking forward to playing with it!

PR #28, opened 2017-02-12
Allow recursive object types

### LINKED ISSUE #27: How to return recursive schema types
In the included example, if we imagined one more field for users that contained their coworkers' ids:

```ml
type user = {
  id   : int;
  name : string;
  role : role;
  ip   : string;
  coworker_ids : int list;
}

let users = [
  { id = 1; name = "Alice"; role = Admin; ip = "unresolved"; coworker_ids = [1]};
  { id = 2; name = "Bob"; role = User; ip = "unresolved"; coworker_ids = [2]};
  { id = 3; name = "Carol"; role = Admin; ip = "unresolved"; coworker_ids = []}
]
```

How would we modify the `user` definition to add a coworkers field?

```ml
let rec user = Schema.(obj
  ~name:"user"
  ~fields:[
      field "coworkers"
      ~args:Arg.[]
      ~typ:(non_null (list (non_null user)))
      ~resolve:(fun () p ->
      let other_ids = p.coworker_ids in
      let others = List.map (fun id -> List.find (fun worker -> worker.id = id) users) other_ids in
      others
      )
  ;
```

Specifically, I get this error with the above code: `Error: This kind of expression is not allowed as right-hand side of 'let rec'`

andreas:
This PR adds the ability to define a recursive object type: an object with a field that refers to the type itself. An example would be a tweet, which has many replies (also tweets).

To achieve this, the fields of an object is now a function that receives the type itself and must return a list of fields. The new type of `obj` is thus:

```ocaml
val obj : name:string ->
          fields:(('ctx, 'src option) typ -> ('ctx, 'src) field list) -> (* <-- new *)
          ('ctx, 'src option) typ
```

Example:

```ocaml
type tweet = {
  id : int;
  replies : tweet list;
}

let tweet = Schema.(obj
  ~name:"tweet"
  ~fields:(fun tweet -> [ (* <-- new *)
    field "id"
      ~typ:(non_null int)
      ~args:Arg.[]
      ~resolver:(fun ctx t -> t.id)
    ;
    field "replies"
      ~typ:(non_null (list tweet))
      ~args:Arg.[]
      ~resolver:(fun ctx t -> t.replies)
  ])
)
```

Behind the scenes, this is achieved using `let rec` and `lazy`. Thanks to @Drup for pointing me in the right direction.

closes #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ let role = Schema.enum
 
 let user = Schema.(obj
   ~name:"user"
-  ~fields:[
+  ~fields:(fun _ -> [
     field "id"
       ~typ:(non_null int)
       ~args:Arg.[]
@@ -73,7 +73,7 @@ let user = Schema.(obj
       ~typ:(non_null role)
       ~args:Arg.[]
       ~resolve:(fun () p -> p.role)
-  ]
+  ])
 )
 
 let schema = Schema.(schema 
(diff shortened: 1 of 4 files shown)


Drup:
Nice. ;)

PR #30, opened 2017-02-26
Proper default arguments

### LINKED ISSUE #29: How to pass arguments from one field to the next
This is related to #28, trying to have an argument for a field that allows explicit depth, e.g.

```
item(id: 1) { 
  comments(depth: 10) {
    text
    id
    comments
  }
}
```

Assuming `comment` is an `io_field`,

1. How do I provide a default value for `depth`? `~default:(Some 1)` causes a type error I can't quite get around
2. Each time before recursing into comments, I want to check if `depth` is positive, and if so recurse into `comments` while decrementing the `depth` argument.

Would you mind providing an example?

andreas:
Default arguments doesn't work correctly in `master` right now. In particular, the type of `arg` isn't correct:

```ocaml
val arg : ?default:'a -> string -> ('a, 'b) arg_typ -> ('a, 'b) arg
```

Here `'a` is the "unification variable", and not the type of the content (as describe [here](https://drup.github.io/2016/08/02/difflists/)).

Specifying a default value for an argument only really makes sense if the argument is nullable, so the default value type and the argument type is interdependent. As such, two functions are now provided to construct arguments (with and without a default value):

```ocaml
val arg  : string -> typ:('a, 'b -> 'a) arg_typ -> ('a, 'b -> 'a) arg
val arg' : string -> typ:('a, 'b option -> 'a) arg_typ -> default:'b -> ('a, 'b -> 'a) arg
```

Example:

```ocaml
let schema = Schema.(schema 
    ~fields:[
      field "greeter"
        ~typ:(non_null string)
        ~args:Arg.[
            arg' "greeting" ~typ:string ~default:"hello";
            arg "name" ~typ:(non_null string)
          ]
        ~resolve:(fun ctx () greeting name ->
          Format.sprintf "%s, %s" greeting name
        )
      ;
    ]
)
```

closes #29

### DIFF
--- a/examples/server.ml
+++ b/examples/server.ml
@@ -57,7 +57,7 @@ let schema = Schema.(schema
         ~typ:string
         ~args:Arg.[
           arg "config" ~typ:(non_null (obj ~name:"greeter_config" ~coerce:(fun greeting name -> (greeting, name)) ~fields:[
-            arg "greeting" ~typ:(non_null string);
+            arg' "greeting" ~typ:string ~default:"hello";
             arg "name" ~typ:(non_null string)
           ]))
         ]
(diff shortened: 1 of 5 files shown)


sgrove:
This seems like a good idea (only slightly concerned about the name, but actually seems alright and I don't have any better suggestions)

PR #45, opened 2017-05-12
Remove depopts and port to jbuilder

rgrinberg:
My WIP port of jbuilder. It runs all the tests at this point. Try it out with:

```
$ make check
```

Note that I split the graphql/lwt/async tests into separate test suites. This is
so that users are able to run tests only for the packages they want to jbuild.

cc @sgrove


andreas:
Thanks for your effort so far! When I run `jbuilder runtest` locally I get the following error:

```
File "graphql/test/parser_test.ml", line 16, characters 17-48:
Error: Unbound value Graphql_parser.sexp_of_document
```

It looks like this issue: https://github.com/janestreet/jbuilder/issues/66. As such, should `ppx_sexp_conv>=0.9` be a requirement?

rgrinberg:
Yes, that's indeed the case. That's the minimum version of ppx_sexp_conv that uses omp. I'll update the constraints.

rgrinberg:
ping? is there interest in this?

graphql-async/test/jbuild
```diff
@@ -0,0 +1,10 @@
+(jbuild_version 1)
+
+(executables
+ ((libraries (graphql_async alcotest async))
```
rgrinberg: It can be Async_kernel I ran into some problems wrt 0.9.0 and 113. Could you upgrade the lib to use the newest async first and then I'll sort this out? Or should I just include the async upgrade in this PR?

andreas: I think it's fine to include #48 in this PR (2a6988ec83a8085079a90e6139b356e919d6f271), otherwise Travis will just fail on that.

rgrinberg: OK this is good now. Note that for async we depend on async_kernel for the dependency itself and async_unis is only required for tests.

andreas:
Thanks a lot for your contribution, @rgrinberg! Merged as part of #50.

PR #56, opened 2017-07-12
graphql-ppx

### LINKED ISSUE #36: GraphQL client
I'm able to use `ocaml-graphql-server` to build a demo app with the Apollo-js client, which is great!

I have a lot of the pieces in place to use a full OCaml/GraphQL stack, but the client (actually in reason) is the last challenge here. Seems like code-generation is the way to go, for lots of reasons - it can be type-checked (both client and server-side), optimized by the server to that only the hash of the query is sent over rather than the full text, etc.

I'd ultimately like to follow the design of the iOS apollo client (https://github.com/apollographql/apollo-ios), but for the time being I'm just looking to get basic query/response working - no pagination, subscriptions, etc.

Seems like the best way to do this is to have GraphQL files and a build step that takes them + a running server url and generates an OCaml module for each one, with its own `executeRequest`, `executeRequestByHash`, etc. methods.

How would you recommend going about this? Is there some way to easily parse a GraphQL query-text and then introspect a schema and its types so that I can pattern match down it and emit a Reason file?

andreas:
This PR is the initial take on `graphql-ppx`, a PPX extension for writing type-safe GraphQL queries. Given a GraphQL schema (introspection query response) and a GraphQL query, `graphql-ppx` generates three values: (1) a GraphQL query (2) a function to construct the associated query variables, and (3) a function for parsing the GraphQL JSON response into a typed value (object type).

My current thinking is that `graphql-ppx` provides the foundation for creating a great GraphQL client experience, but will not include that itself. Rather, I imagine a separate`graphql-client` library building on top of e.g. `cohttp` and `graphql-ppx`.

Here's an example of using `graphql-ppx` and the generated values (the schema is shown at the top in [GraphQL Schema Language](https://raw.githubusercontent.com/sogko/graphql-shorthand-notation-cheat-sheet/master/graphql-shorthand-notation-cheat-sheet.png)):

```ocaml
(*
enum ROLE {
  USER
  ADMIN
}

type User {
  id: ID!
  role: ROLE
  contacts: [User!]!
}

type Query {
  user(id: ID!): User
}

schema {
  query: Query
}
*)

let query, kvariables, parse = [%graphql {|
  query FindUser($id: ID!) {
    user(id: $id) {
      id
      role
      contacts {
        id
      }
    }
  }
|}] in
(* ... *)
```

In this example, the following values are generated:

- `query` (type `string`) is the GraphQL query to be submitted. Currently it's an unmodified version of the string provided to `%graphql`, but it will likely be modified in the future, e.g. to inject `__typename` for interface disambiguation.
- `kvariables` (type `(Yojson.Basic.json -> 'a) -> id:string -> unit -> 'a`) is a function to construct the JSON value to submit as query variables ([doc](http://graphql.org/learn/serving-over-http/#post-request)). Note that the first argument is a continuation to handle the resulting JSON value -- this makes it easier to write nice clients (see more below). The type is extracted from the query. Required variables appear as labeled arguments, optional variables appear as optional arguments.
- `parse` is a function for parsing the JSON response from the server and has the type:
  
  ```
  Yojson.Basic.json ->
    <user:
      <id: string;
      role: [> `USER | `ADMIN] option;
      contacts: <id: string> list>
    >
  ```
  This type captures the shape of the GraphQL response in a type-safe fashion based on the provided schema. Scalars are converted to their OCaml equivalent (e.g. a GraphQL `String` is an OCaml `string`), nullable types are converted to `option` types, enums to polymorphic variants, lists to list types and GraphQL objects to OCaml objects. Note that this function will likely return a `result` type in the future, as the GraphQL query can fail.

With the above, it's possible to write quite executable queries quite easily:

```ocaml
let executable_query (query, kvariables, parse) =
  kvariables (fun variables ->
    let response_body = (* construct HTTP body here and submit to GraphQL endpoint *) in
    Yojson.Basic.of_string response_body
    |> parse
  )

let find_user_role = executable_query [%graphql {|
  query FindUserRole($id: ID!) {
    user(id: $id) {
      role
    }
  }
|}]
```
Here  `find_user_role` has the type ```id:string -> unit -> <user: <role: [`USER | `ADMIN] option> option>```. Quite nice 😄 See [`github.ml`](https://github.com/andreas/ocaml-graphql-server/blob/ppx/graphql-ppx/examples/github.ml) for a real example using `Lwt` and `Cohttp`.

With all this being said, this is still very early, and there are a number of limitations:

- No support for input objects
- No support for unions
- No support for interfaces
- No support for custom scalar types
- Poor error handling
- Error reporting should be improved
- Path to JSON introspection query result is hardcoded to "schema.json"
- Assumes the query has already been validated

Input is very welcome!

resolves #36


sgrove:
Looks very cool indeed!

jaredly:
wild idea -- what if we used straight ocaml/reason syntax instead of parsing a string?

```
[%graphql
let findUser (id: id) => {
  user: fun ::id=id => {
    id,
    role,
    contacts: {
      id,
    }
  }
}]
```

We could potentially get syntax highlighting, type hinting, nice error messages etc. if we set it up right.

neonsquare:
@jaredly 
Hm... unconvincing

+ Syntax highlighting and completion on whatever is the current reason syntax ( ;-P )

- Cannot just use GraphQL query examples by copy paste
- Cannot use editor highlighting/tools that know GraphQL Query syntax
- Cannot use infrastructure to externalize the query documents
- External query analyzers need to know Reason syntax

andreas:
@jaredly: I really like the idea of type hinting -- creating a GraphiQL-like experience within your editor would be amazing. I agree with the points made by @neonsquare though, and would prefer if we can keep regular GraphQL syntax. Maybe Merlin could support type hints from PPX syntax extensions?

jaredly:
I think the blocker would be that the editor has to be able to tokenize the multiline string in order for merlin to know what coordinates to ask about.
Merlin does support type hinting from ppx syntax extensions, but only if the contents are valid ocaml syntax (as my idea demonstrated).
It is nice to be able to just drop in a graphql example and have it work, though. I was just hoping to avoid more layers than necessary, but maybe it's not practical.

andreas:
I've changed the generated variable function to be continuation passing style, i.e. it now accepts a function `(Yojson.Basic.json -> 'a) -> arg1:int -> arg2:string option -> 'a`. This makes it really easy to make queries executable and retain a nice type.

I've added a section in the PR description which details this, and added an real-life example of fetching a repo from Github: [`github.ml`](https://github.com/andreas/ocaml-graphql-server/blob/ppx/graphql-ppx/examples/github.ml).

wokalski:
would it be possible to expose a Functor so that the JSON parser is composable? It would be tremendous to use this ppx with Bucklescript.

andreas:
@wokalski That's an interesting thought! Which other JSON library would you like to use? Do you have a suggested signature?

arnarthor:
I've thought about this a bit myself. Haven't formalised how I would go about creating what I've envisioned yet, and it actually goes way further than just a ppx for graphql, but here goes.

I've been thinking about GraphQL clients in Reason/OCaml, and what I thought could work or at least be a nice experiment is to do like @wokalski says, exposing a functor that allows me to compose JSON parser and have a built in Query method on the returned module.

So in a dreamworld, having something as simple as

```ocaml
module GraphQLForUser = {
  type t;
  let graphqlQuery = [%graphql {|
      query FindUser($id: ID!) {
        user(id: $id) {
          id
          role
          contacts {
            id
          }
        }
      }
    |}];
}
let graphqlForUser = GraphQL(GraphQLForUser);
graphqlForUser.query |> parse |> <use data here>
```

Disclaimer, I haven't put a ton of thought into this, just been wondering if something like this could be achieved by utilising the typesystem.

This could allow us to create simple clients all over the place, and type the type of GraphQL response we want in components/functions by saying the parameter should be of type `GraphQLForUser.t`


Sorry for the caveat of this PR 😄 just wanted to express my thoughts on this. Love this initiative, it's definitely a great start to get first class citizen GraphQL into Reason/OCaml

wokalski:
@arnarthor I don't think I understand how it is more powerful than the ppx. 


@andreas In any case I don't have a favourite Json library in OCaml. I'm more thinking about Js where I use `bs-json` which uses JS platform functions behind the scenes.

arnarthor:
Definitely not more powerful than ppx, it could use the ppx to generate the query itself. This is just my vision of how it could be nice to create GraphQL clients for different queries by wrapping everything into a module that uses fetch (or some other http client) behind the scenes to do the request for you.

andreas:
@arnarthor Your example seems very close to what is already available, only it's wrapped in some module/functors. What do you see as the pros/cons compared to the proposed solution which does not used modules?

Here's your example in the proposed style:

```ocaml
let graphql_for_user = executable_query [%graphql {|
  query FindUser($id: ID!) {
    user(id: $id) {
      id
      role
      contacts {
        id
      }
    }
  }
|}]

let user_record = graphql_for_user ~id:"42" ()
```

where `executable_query` defined similarly to [this](https://github.com/andreas/ocaml-graphql-server/blob/cdaa16cca89175f026f53378ebb8bdd1b590a3ea/graphql-ppx/examples/github.ml#L3-L26).

@wokalski That makes sense. The type `Yojson.Basic.json` is transparent, so it should be possible to write an adaptor to `bs-json`. Functorizing `graphql-ppx` is also a possibility, in which case we would have to find a signature that palatable to most JSON libraries. It would complicate `graphql-ppx`, but I do see the value in good Bucklescript support.

wokalski:
If `Yojson` is the de-facto standard when compiling to native, and `Js.Json`/`bs-json` is the standard on Bucklescript, there's little (to no) value in providing additional functionality which will never be utilised. There's also `jsoo` which is probably slightly different in this aspect but I've never used it.

wokalski:
@andreas I went through the PR and it looks like using `bs-json` if there's using a functor will be super simple. It has a very similar interface to yosjon with some exceptions.

andreas:
@wokalski I'm currently considering the following options for supporting different JSON types:

- Assume a particular module `Json` with a certain signature is in scope for constructing/parsing JSON values. This is similar to [`ppx_let`](https://github.com/janestreet/ppx_let#getting-the-right-names-in-scope).
- Annotate the GraphQL query string with the name of a module with an assumed particular signature `[%graphql "..."[@json Yojson]]`.
- Expose an identifier per JSON type, e.g. `[%graphql.yojson "..."]`, `[%graphql.bs "..."]`, etc. This scheme would not be extensible.

Alternatively `graphql-ppx` could possibly detect whether it's compiling for native or Bucklescript and automatically use `yojson` or `bs-json` -- though maybe that's too magic?

wokalski:
I think the second option is the least magical and the amount of boilerplate is negligible. It's also very explicit. I love it.

andreas:
I ended up moving this to a separate repo: [`ppx_graphql`](https://github.com/andreas/ppx_graphql).

PR #62, opened 2017-09-05
Parse tab as whitespace

jagare:
According to GraphQL spec tab should be treated as whitespace. Without this fix there is a parse error if a tab is inserted in a query.


andreas:
Thanks @jagare!

PR #64, opened 2017-09-17
Parser fixes

### LINKED ISSUE #68: Slash \ escaped characters in direct input strings aren't unescaped
If I run this query in GraphiQL

```graphql
query {
  goomba(alias: "test\"") {
    id: ID!
  }
}
```

the value passed to the resolve function will be exactly `test\"`. However, if I change the query to

```graphql
query GoombaQuery($input:String!) {
  goomba(alias: $input) {
    id: ID!
  }
}
```

with the variables

```json
{ "input": "test\"" }
```

the quotation is de-escaped as expected, `test"`.

andreas:
- Depend on `angstrom` 0.7.0: https://github.com/ocaml/opam-repository/pull/10286/#issuecomment-329967555
- Fix parsing of escaped characters

closes #68

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@
 
 - Parse tabs as whitespace (#62)
 - Move parser to separate package (#63)
+- Depend on angstrom 0.7.0 (#64)
+- Fix parsing of quoted strings (#64)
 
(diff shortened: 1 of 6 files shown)


andreas:
Thanks! Fixed.

Issue #61, opened 2017-08-20
Propagating errors along with data


dwwoelfel:
Is there a way for the resolver functions to propagate errors? Ideally, each resolver could return both data and errors, which would be collected at the top level with information about where the error occurred (https://facebook.github.io/graphql/#sec-Errors). 

Right now, we're throwing exceptions from the resolvers, but this means that 1. we can't notify the caller about non-fatal errors, and 2. the graphql stack can't add the `path` to the error when the error is specific to a certain field. 

We'd like to return something like this to the client:

```
{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ]
    }
  ],
  "data": {
    "hero": {
      "name": "R2-D2",
      "heroFriends": [
        {
          "id": "1000",
          "name": "Luke Skywalker"
        },
        {
          "id": "1002",
          "name": null
        },
        {
          "id": "1003",
          "name": "Leia Organa"
        }
      ]
    }
  }
}
```

andreas:
The story around error handling is indeed lacking right now. As you mention, a resolver should be able indicate an error occurred, and this error should propagate up until a nullable ancestor is found. I'm currently leaning towards an exception-based approach, as I fear enforcing all resolvers to return a result-type is too cumbersome. I imagine it would be somewhat like [monitors in Async](https://realworldocaml.org/v1/en/html/concurrent-programming-with-async.html#monitors).

donut:
As it stands, is there any way to respond with error messaging other than "Error: Internal Server Error"?  Is the only option to create wrapper types that have optional error/value properties? Something like,

```graphql
Book {
  id: ID!
  title: String!
  author: String!
}

Error {
  code: Int!
  message: String!
}

BookOrError {
  error: Error
  book: Book
}
```

Just want to confirm before I build all these wrapper types.

andreas:
@donut If you want a short-term solution (aka a hack!), I would suggest:

1) Use a map keyed with a unique id per request to store errors.
2) If an error occurs in a `resolve`-function, return `None` and put an error in that map.
3) In the `callback`-function to `Cohttp.Server.create`, add errors to the GraphQL response from the error map and clear out the request id.

[...]

PR #69, opened 2017-10-23
Error handling

andreas:
This PR changes the type of `Schema.io_field`, such that the `resolve`-function must return a `('a, string) result`:

```ocaml
(* before *)
val io_field : ?doc:string ->
               ?deprecated:deprecated ->
               string ->
               typ:('ctx, 'out) typ ->
               args:('out io, 'args) Arg.arg_list ->
               resolve:('ctx -> 'src -> 'args) ->
               ('ctx, 'src) field

(* after *)
val io_field : ?doc:string ->
               ?deprecated:deprecated ->
               string ->
               typ:('ctx, 'out) typ ->
               args:(('out, string) result io, 'args) Arg.arg_list ->
               resolve:('ctx -> 'src -> 'args) ->
               ('ctx, 'src) field
```

Errors returned from `resolve`-functions will be exposed in the `errors`-part of the response. Nullable fields that return `Error` will be `null`, while errors from non-nullable fields will bubble up as [described in the spec](http://facebook.github.io/graphql/October2016/#sec-Errors-and-Non-Nullability).

This is a breaking change.

Addresses #61

### DIFF
--- a/graphql-async/test/async_test.ml
+++ b/graphql-async/test/async_test.ml
@@ -33,7 +33,7 @@ let schema = Graphql_async.Schema.(schema [
       io_field "io_int"
         ~typ:(non_null int)
         ~args:Arg.[]
-        ~resolve:(fun () () -> Deferred.return 42)
+        ~resolve:(fun () () -> Deferred.return (Ok 42))
     ]
 )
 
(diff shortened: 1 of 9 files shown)


andreas:
Yeah, I've postponed merging this PR for a long time while trying to come up with a design which unifies middleware, error handling and more arguments for resolve-functions. Alas I think it's time to move forward, and possibly accept another breaking change if such a design emerges later.

Thanks for the nudge 👍

PR #70, opened 2017-11-30
Union and interface support

### LINKED ISSUE #6: Interface support
See:

- https://facebook.github.io/graphql/#sec-Interfaces
- https://facebook.github.io/graphql/#sec-Interface
- https://facebook.github.io/graphql/#sec-Field-Selections-on-Objects-Interfaces-and-Unions-Types

andreas:
This PR adds support for GraphQL union and interface types.

The examples below assume the following types and values:

```ocaml
type dog
val fido : dog
val dog : ('ctx, dog option) Schema.typ

type cat
val meow : cat
val cat : ('ctx, cat option) Schema.typ
```

## Union

Unions are constructed with `Schema.union` and requires a type annotation to distinguish between other unions and interfaces (this is similar to [`ctypes`](https://github.com/ocamllabs/ocaml-ctypes/issues/322)). Adding a object type `('ctx, 'src option) typ` to a union type `('ctx, ('ctx, 'a) abstract_value) typ` returns a function to embed into the union type: `'src -> ('ctx, 'a) abstract_value`.

```ocaml
let pet : ('ctx, [`Pet]) Schema.abstract_typ = Schema.union "Pet"
let cat_as_pet = Schema.add_typ pet cat
let dog_as_pet = Schema.add_typ pet dog

Schema.(schema [
  field "pets"
    ~typ:(list (non_null pet))
    ~args:Arg.[]
    ~resolve:(fun ctx () -> Some [cat_as_pet meow; dog_as_pet fido])
])
```

## Interface

Interfaces work very similar to unions. Interfaces are constructed with `Schema.interface` and requires a type annotation to distinguish between unions and interfaces. Adding a object type `('ctx, 'src option) typ` to an interface type `('ctx, ('ctx, 'a) abstract_value) typ` returns a function to embed into the union type: `'src -> ('ctx, 'a) abstract_value`.

```ocaml
let nameable : ('ctx, [`Nameable]) Schema.abstract_typ = Schema.interface "Nameable"
  ~fields:(fun _ -> [
    abstract_field "name"
      ~typ:(non_null string)
      ~args:Arg.[]
  ])

(* raises if cat does not implement Nameable *)
let cat_as_nameable = Schema.add_typ nameable cat

(* raises if dog does not implement Nameable *)
let dog_as_nameable = Schema.add_typ nameable dog 

Schema.(schema [
  field "nameable"
    ~typ:(list (non_null nameable))
    ~args:Arg.[]
    ~resolve:(fun ctx () -> Some [cat_as_nameable meow; dog_as_nameable fido])
])
```

## Future

- Currently a subtyping check is missing when adding a type to an interface. This will yield a runtime exception when calling `Schema.add_type`. Although inconvenient, I think it's a reasonable trade-off to have an exception at schema-construction time to be exception-free at query-execution time.
- It may be possible to get rid of the manual type ascription using generative functors, similarly to what is described for [`ctypes`](https://github.com/ocamllabs/ocaml-ctypes/issues/322).

Closes #6, #7


shinzui:
Why isn't this merged?

andreas:
@shinzui I wanted to get rid of the required type annotation as described in the PR. I did make some progress with this, but the result was not substantially better. Essentially it ended up looking like this:

```ocaml
module Pet = (val Schema.union "Pet" : Schema.Abstract with type ctx = unit)
let cat_as_pet = Pet.add_type cat
let dog_as_pet = Pet.add_type dog
```

I haven't found a way to avoid the type annotation for `Pet`, unfortunately.

Finally, there was a remaining bug regarding nested fragment spreads and abstract types ([test case](https://github.com/andreas/ocaml-graphql-server/blob/013bfc6ec27233f2e9a605160cfc3cb0b9732426/graphql/test/abstract_test.ml#L150)). That should have been resolved now.

PR #72, opened 2017-12-06
Make custom argument types generalizable

andreas:
Previously argument types, `Schema.Arg.arg_typ`, was not generalizable, i.e. they could not be used in multiple contexts. Example that didn't work prior to this PR:

```ocaml
let silly_int = Schema.Arg.scalar "SillyInt"
  ~coerce:(function
    | `Int n -> Ok (n+1)
    | _ -> Error "Invalid silly int"
  )

let schema = Schema.(schema [
  field "foo"
    ~typ:(non_null int)
    ~args:Arg.[arg "a" ~typ:(non_null silly_int)]
    ~resolve:(fun _ () a -> a)
  ;
  field "bar"
    ~typ:(non_null string)
    ~args:Arg.[arg "a" ~typ:(non_null silly_int)]  (* <-- this would fail! *)
    ~resolve:(fun _ () a -> string_of_int a)
])
```

This was an artifact of the clumsy way `Schema.Arg.arg_typ` was defined. With this PR, the above example works. Thanks to @drup for the suggested simplification.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - Move parser to separate package (#63)
 - Depend on angstrom 0.7.0 (#64)
 - Fix parsing of quoted strings (#64)
+- Make custom argument types generalizable (#65)
 
 0.3.0 2017-08-31
 ---------------------------------
--- a/graphql/src/graphql_intf.ml
+++ b/graphql/src/graphql_intf.ml
@@ -36,48 +36,48 @@ module type Schema = sig
             ('ctx, 'src option) typ
 
   module Arg : sig
-    type (_, _) arg
-    type (_, _) arg_typ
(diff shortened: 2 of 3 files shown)


Drup:
Looks good, well done! :)

kennetpostigo:
What did you fix that `not generalizable`? I see this occasionally, wondering how you get around that

Drup:
@kennetpostigo Generalization (or lack thereof) is caused by (weak) polymorphism. In this case, the solution was simply to remove the polymorphism when it was not needed. the `arg` type used to be of the form `('a, T -> 'a) arg`. Now it's simply `T arg` with T a concrete type (int, string, ...).

No more polymorphism, no more problems. :)

PR #80, opened 2018-02-10
Replace sexplib with fmt for Graphql_parser

andreas:
This PR replaces `sexplib` with `fmt` for `graphql_parser`. This means you can no longer convert a value of type `Graphql_parser.document` to an s-expression, but instead you can convert it back to a string representing the original GraphQL query. Note that formatting and comments are lost in this conversion. This ability also simplifies the tests, which are now simply of the form `to_string(parse(query)) == query`. Finally, `fmt` is a much lighter dependency than `sexplib`.


andreas:
Superceded by #132

Issue #84, opened 2018-03-21
Error in the example


mnewsha:
Hi,
I got this error trying to build the example:
```
File "server.ml", line 73, characters 2-14:
Error: Unbound module Server
```
What is wrong?

grsabreu:
did you run jbuilder?

mnewsha:
yes I ran jbuilder as ```jbuilder build server.exe``` and I got this error.

mnewsha:
I have changed the second line from ```module C = Cohttp_lwt_unix``` to ```open Cohttp_lwt_unix``` and it worked but http://localhost:8080 seems not getting any response on client side

andreas:
@mnewsha, I don't think #85 is not a correct fix. The example compiles correct for me with `graphql-lwt` version `0.5.0`. Could you please check your version of `graphql-lwt` (try `opam info graphql-lwt`)?

mnewsha:
The version of graphql-lwt I'm using right now is 0.1.0 on OCaml 4.04.2 and unfortunately the available versions of the package for this version of OCaml are only 0.1.0 and 0.3.0 do you think upgrading to 0.3.0 can help? it needs some downgradings for cohttp and prometheus packages.
In fact I was trying to use ocaml-graphql-server with Ocsigen and I wrote this piece of code in a file ```message.eliom```:
```
open Graphql

type message = {
  msg : string;
}

let messages = [
  {msg = "Hello!"};
  {msg = "Bonjour!"};
  {msg = "Buongiorno!"}
]

let message = Schema.(obj "message"
  ~doc: "Message to be shown to user"
  ~fields: (fun _ -> [
    field "msg"
      ~doc: "Greeting for example"
      ~typ: (non_null string)
      ~args: Arg.[]
      ~resolve: (fun ctx c -> c.msg)
  ])
)

let schema = Schema.(schema [
  field "mssages"
    ~typ: (non_null (list (non_null message)))
    ~args: Arg.[]
    ~resolve:(fun ctx () -> messages)
])
```
I had this idea to use a part of the example on main page:
```
let query = Graphql_parser.parse "{ users(limit: $x) { name } }" in
let json_variables = Yojson.Basic.(from_string "{\"x\": 42}" |> Util.to_assoc) in
let variables = (json_variables :> (string * Graphql_parser.const_value) list)
Graphql.Schema.execute schema ctx ~variables query
```
[...]

PR #85, opened 2018-03-22
Lets example be compiled

mnewsha:
Eliminate the error:
File "server.ml", line 73, characters 2-14:
Error: Unbound module Server

### DIFF
--- a/examples/server.ml
+++ b/examples/server.ml
@@ -1,5 +1,5 @@
 open Lwt
-module C = Cohttp_lwt_unix
+open Cohttp_lwt_unix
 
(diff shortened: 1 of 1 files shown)


andreas:
As mentioned in #84, I don't think this is a correct fix. Let's continue the discussion there.

guilhermedecampo:
Yes, this doesn't fix the issue.

screen shot 2018-03-28 at 15 12 36

andreas:
@guilhermedecampo, as mentioned in #84, please check you're using `graphql-lwt` version `0.3.0`.

PR #87, opened 2018-04-01
Add cohttp-lwt-unix as examples' dune dependecy

turboMaCk:
**NOOB ALERT: This might be pretty stupid change as I don't have almost any experience with OCaml ecosystem. If it's the case - close this.**

Cohttp_lwt_unix is dependency used in examples but was missing in
jbuild file. This commit should resolve potential issues with
unbounded modules while attempting to build examples on systems on
which this lib wasn't already installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ To run a sample GraphQL server also serving GraphiQL, do the following:
 
 ```bash
 opam install graphql-lwt jbuilder
-git checkout git@github.com/andreas/ocaml-graphql-server.git
+git clone git@github.com/andreas/ocaml-graphql-server.git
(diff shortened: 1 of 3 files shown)


andreas:
Thanks for reporting this 😄 

I assume the offending line is [`examples/server.ml:2`](https://github.com/andreas/ocaml-graphql-server/blob/2bac77fe1e5324f903d69ca7578341df3dbf9ca9/examples/server.ml#L2)? That line is actually not used, and is a leftover from a refactoring -- I think it can be removed.

turboMaCk:
I see. The correct fix would be probably then just about removing that import.

PR #91, opened 2018-05-22
Support GET requests

### LINKED ISSUE #89: GraphiQL returned for valid GET queries
http://graphql.org/learn/serving-over-http/#get-request

Queries like this should work:

```
curl 'localhost:8080/graphql?query=%examples%7Bid%7D%7'
```

But instead I get the GraphiQL HTML page.

I think we need to make two changes:

 - Support reading the query from query params
 - Check accept headers for `text/html` before sending GraphiQL

brendanlong:
Previously we were serving the GraphiQL page for all GET requests, but
we should only do that if we receive `Accept: text/html`.

We also ignored the `query` query param, even though it's the required
way to handle GET requests, and is recommended to be supported for
POST requests.

See: http://graphql.org/learn/serving-over-http/#get-request

Fixes #89

### DIFF
--- a/graphql-lwt/src/graphql_lwt.ml
+++ b/graphql-lwt/src/graphql_lwt.ml
@@ -21,11 +21,27 @@ module Server = struct
     Schema.execute schema ctx ~variables doc
 
   let execute_request ctx schema req body =
-    Cohttp_lwt.Body.to_string body >>= fun body' ->
-    Lwt_io.printf "Body: %s\n" body';
-    let json = Yojson.Basic.from_string body' in
-    let query = Yojson.Basic.(json |> Util.member "query" |> Util.to_string) in
-    let variables = try Yojson.Basic.Util.(json |> member "variables" |> to_assoc) with _ -> [] in
+    let uri = Cohttp.Request.uri req in
+    (* Use query params if the "query" query param exists; otherwise use the
+       body of the request *)
+    (match Uri.get_query_param uri "query" with
+      | Some query ->
+        Cohttp_lwt.Body.drain_body body >|= fun () ->
+        let variables =
+          match Uri.get_query_param uri "variables" with
+          | Some json ->
+            Yojson.Basic.(from_string json |> Util.to_assoc)
+          | None -> []
+        in
+        query, variables
+      | None ->
+        Cohttp_lwt.Body.to_string body >|= fun body' ->
+        Lwt_io.printf "Body: %s\n" body';
+        let json = Yojson.Basic.from_string body' in
+        let query = Yojson.Basic.(json |> Util.member "query" |> Util.to_string) in
+        let variables = try Yojson.Basic.Util.(json |> member "variables" |> to_assoc) with _ -> [] in
+        query, variables)
+    >>= fun (query, variables) ->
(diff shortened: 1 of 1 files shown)


brendanlong:
Sorry if the formatting is weird. I was having trouble understanding the current pattern, especially that indented match request at the end of `mk_callback`. If you have a style guide to point at I can make the code follow that.

graphql-lwt/src/graphql_lwt.ml
```diff
@@ -21,11 +21,27 @@ module Server = struct
...
+          | None -> []
+        in
+        query, variables
+      | None ->
+        Cohttp_lwt.Body.to_string body >|= fun body' ->
+        Lwt_io.printf "Body: %s\n" body';
+        let json = Yojson.Basic.from_string body' in
+        let query = Yojson.Basic.(json |> Util.member "query" |> Util.to_string) in
+        let variables = try Yojson.Basic.Util.(json |> member "variables" |> to_assoc) with _ -> [] in
+        query, variables)
```
andreas: Maybe this `match` could be extracted to a helper function `Uri.t -> Cohttp_lwt.Body.t -> string * Yojson.Basic.json`?

graphql-lwt/src/graphql_lwt.ml
```diff
@@ -40,10 +56,20 @@ module Server = struct
     Lwt_io.printf "Req: %s\n" req.resource;
     let req_path = Cohttp.Request.uri req |> Uri.path in
     let path_parts = Str.(split (regexp "/") req_path) in
-      match req.meth, path_parts with
-      | `GET,  ["graphql"]       -> static_file_response "index.html"
-      | `GET,  ["graphql"; path] -> static_file_response path
-      | `POST, ["graphql"]       -> execute_request (mk_context req) schema req body
+    let accept_html =
+      let headers = Cohttp.Request.headers req in
+      let accept = Cohttp.Header.get headers "accept" in
```
andreas: Could this use `Cohttp.Header.get_multi` instead?

graphql-lwt/src/graphql_lwt.ml
```diff
@@ -21,11 +21,27 @@ module Server = struct
...
   let execute_request ctx schema req body =
-    Cohttp_lwt.Body.to_string body >>= fun body' ->
-    Lwt_io.printf "Body: %s\n" body';
-    let json = Yojson.Basic.from_string body' in
-    let query = Yojson.Basic.(json |> Util.member "query" |> Util.to_string) in
-    let variables = try Yojson.Basic.Util.(json |> member "variables" |> to_assoc) with _ -> [] in
+    let uri = Cohttp.Request.uri req in
+    (* Use query params if the "query" query param exists; otherwise use the
+       body of the request *)
+    (match Uri.get_query_param uri "query" with
```
andreas: What are your thoughts on checking `req.meth` rather than whether `query` is present as a query parameter? I.e. when doing a POST to `/graphql?query=...`.

andreas: Good point! Would be great to support all the recommended behaviours.

andreas:
Superceded by https://github.com/andreas/ocaml-graphql-server/pull/156.

PR #94, opened 2018-07-05
Abstract types: Update interface list

osener:
With this change, GraphiQL can list implementations of interfaces:

image

### DIFF
--- a/graphql/src/graphql_schema.ml
+++ b/graphql/src/graphql_schema.ml
@@ -387,12 +387,15 @@ module Make(Io : IO) = struct
     i
 
   let add_type abstract_typ typ =
-    match abstract_typ with
-    | Abstract a ->
+    match (abstract_typ, typ) with
+    | Abstract ({ kind = `Interface _; _ } as a), Object o ->
         (* TODO add subtype check here *)
         a.types <- (AnyTyp typ)::a.types;
-        fun src ->
-          AbstractValue (typ, src)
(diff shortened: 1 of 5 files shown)


graphql/src/graphql_schema.ml
```diff
@@ -391,6 +391,9 @@ module Make(Io : IO) = struct
     | Abstract a ->
         (* TODO add subtype check here *)
         a.types <- (AnyTyp typ)::a.types;
+        (match (a, typ) with
+         | ({ kind= `Interface _; _}, Object o) -> o.interfaces := a :: !(o.interfaces)
+         | _ -> ());
```
andreas: How about matching on both `abstract_typ` and `typ` in the outer `match`?

```ocaml
match abstract_typ, typ with
| Abstract ({ kind = `Interface _; _ } as a), Object o ->
    (* TODO add subtype check here *)
    a.types <- (AnyTyp typ)::a.types;
    o.interfaces := a :: !o.interfaces
| Abstract ({ kind = `Union; _ } as a), Object o ->
    a.types <- (AnyTyp typ)::a.types
| _ ->
    failwith "..."
```

osener:
@andreas done! I added a test checking the introspection query result for both unions and interfaces. Let me know if you had something more specific in mind.

andreas:
Thanks for the test case! I ended up rolling these changes into #70 and adapted them slightly to fix another issue. I appreciate the contribution! 👍

PR #95, opened 2018-07-05
Graphql_lwt: Return errors as JSON objects with 200 status code

osener:
With this commit:

- Errors get reported with 200 status code 
GraphQL spec has no mention of error codes since it doesn't care about the transport and uses the `error` field in the response as the source of truth for errors. If I'm understanding the common practice correctly, clients only expect HTTP error codes when the GraphQL server cannot answer the query at all.

  On top of this, GraphQL servers are also expected to return both `data` and `error` fields when it can resolve some of the queries but not the others, making the current 500 code too broad.

  Some references:
    https://github.com/graphql-python/graphene/issues/142
    https://platform.github.community/t/http-status-codes-for-requests/1860

- Errors are returned as a JSON object
  Cohttp's respond_error utility adds the "Error:" prefix to the response, making it harder for the client to parse the result.

### Before (status 500)
screenshot from 2018-07-05 16-56-42

### After (status 200)
screenshot from 2018-07-05 16-56-00

## Next steps
I'd love to have more detailed error responses. For example in Github (image below):
- In the query response, fields that have failed get set to `null`, instead of the catch-all `data: null`
- Errors have things like error type and more information regarding which query they came from.

Apollo's node server has also added something similar recently: https://www.apollographql.com/docs/apollo-server/v2/features/errors.html

screenshot from 2018-07-05 16-57-43

I'll keep investigating this as far as my ocaml-fu allows.

### DIFF
--- a/graphql-lwt/src/graphql_lwt.ml
+++ b/graphql-lwt/src/graphql_lwt.ml
@@ -29,12 +29,9 @@ module Server = struct
     Lwt_io.printf "Query: %s\n" query;
     let result = execute_query ctx schema (variables :> (string * Graphql_parser.const_value) list) query in
     result >>= function
-    | Ok data ->
+    | Ok data | Error data ->
         let body = Yojson.Basic.to_string data in
         C.Server.respond_string ~status:`OK ~body ()
-    | Error err ->
-        let body = Yojson.Basic.to_string err in
-        C.Server.respond_error ~body ()
 
   let mk_callback mk_context schema conn (req : Cohttp.Request.t) body =
     Lwt_io.printf "Req: %s\n" req.resource;


andreas:
Thanks @osener, these are some great observations that should definitely be addressed!

Regarding the status code: always returning `500` on an error indeed seems too simplistic, and the prefix added by `respond_error` is a mistake 😢However, from glancing at the implementation of [`express-graphql`](https://github.com/graphql/express-graphql), it appears always returning 200 is too simplistic also. For example, `express-graphql` returns 400 for syntax errors, and 500 when missing an operation name. Right now, `Schema.execute` does not return very granular errors, but it [does keep track of them internally](https://github.com/andreas/ocaml-graphql-server/blob/bd121b2f8edd53ccaa4a6e8b8377c8e85cd40935/graphql/src/graphql_schema.ml#L1132-L1147). To address this properly, I think we would have to expose these to allow status codes that match `express-graphql`.

> In the query response, fields that have failed get set to null, instead of the catch-all data: null

This should already be the case if you have a nullable field. Can you provide more details?

> Errors have things like error type and more information regarding which query they came from.

This would indeed be a great addition, in particular adding `path`, which is now part of the spec.

osener:
> This should already be the case if you have a nullable field. Can you provide more details?
You're right, it was because I was returning the error from a non-nullable field. Thanks for the pointer!

Would you like me to update this PR to call `respond_string ~status:`Internal_server_error` instead? That would at least allow detecting errors in the result.

andreas:
> Would you like me to update this PR to call respond_string ~status:Internal_server_error instead? That would at least allow detecting errors in the result.

That would be a great first step and immediately useful 👍

osener:
I've opened #96, and I'm closing this PR since changes to the status code needs more thinking than simply returning 200.

PR #96, opened 2018-07-07
Graphql_lwt: Return errors as JSON object

osener:
As discussed in #95

### DIFF
--- a/graphql-lwt/src/graphql_lwt.ml
+++ b/graphql-lwt/src/graphql_lwt.ml
@@ -34,7 +34,7 @@ module Server = struct
         C.Server.respond_string ~status:`OK ~body ()
     | Error err ->
         let body = Yojson.Basic.to_string err in
-        C.Server.respond_error ~body ()
+        C.Server.respond_string ~status:`Internal_server_error ~body ()
 
   let mk_callback mk_context schema conn (req : Cohttp.Request.t) body =
     Lwt_io.printf "Req: %s\n" req.resource;


graphql-lwt/src/graphql_lwt.ml
```diff
@@ -34,7 +34,7 @@ module Server = struct
         C.Server.respond_string ~status:`OK ~body ()
     | Error err ->
         let body = Yojson.Basic.to_string err in
-        C.Server.respond_error ~body ()
+        C.Server.respond_string ~status:`Internal_server_error ~body ()
```
osener: should these also add a `Content-Type` header?

andreas: That would be nice, but I'm fine leaving that for a separate PR.

graphql-lwt/test/lwt_test.ml
```diff
@@ -45,6 +50,15 @@ let suite = [
...
   );
+  ("error reporting", `Quick, fun () ->
+    test_query schema () "{ io_error }" (`Assoc [
+      "data", `Assoc [
+        "io_error", `Null
+      ];
+      "errors", `List [
+        `Assoc [ "message", `String "error" ]
+      ]
+    ]))
```
andreas: I'm not sure this test case adds much beyond what's already covered in [`error_test.ml`](https://github.com/osener/ocaml-graphql-server/blob/dddb241fb679a18605ff66f6ef650d93bb71e7b0/graphql/test/error_test.ml).

It would be great to start a new suite of tests for `Graphql_lwt.Server` though, which is currently untested. I don't consider that a requirement for merging this PR though.

osener: Since this bug happened in graphql-lwt, I thought it would be nice to add a test that checks if errors are returned as objects. Thanks for the pointer to the `error_test` module, I missed that it does test this exact behavior. I'll remove the test from the commit.

andreas:
Thanks!

PR #97, opened 2018-07-10
Fixed git clone command

happylinks:
Not sure if this was intentional, but the checkout didn't work for me.


andreas:
Not intentional at all, just a silly mistake. Thanks for the fix!

PR #65, opened 2017-09-17
 Rewrite parser with Menhir

andreas:
This PR experimentally replaces `angstrom` with Menhir for parsing.

PR #98, opened 2018-07-20
Fix parsing enums that start with e

dwwoelfel:
The parser thinks that enums starting with "E" are numbers because it sees the E as the exponent in a float (e.g. 1e6). 

This PR fixes that problem, but doesn't look like the ideal solution. I'm not too familiar with Angstrom, but it seems like it should be possible to implement something whose logic looks very similar to the spec: http://facebook.github.io/graphql/draft/#sec-Int-Value

Have you considered using libgraphqlparser instead of a custom parser? That could give users of the ocaml-graphql-server a much better experience with parser errors.

### DIFF
--- a/graphql_parser/src/graphql_parser.ml
+++ b/graphql_parser/src/graphql_parser.ml
@@ -158,7 +158,21 @@ let name = take_while1 is_name_char
 
 let is_number_char =
   function | '0' .. '9' | 'e' | 'E' | '.' | '-' | '+' -> true | _ -> false
-let number_chars = take_while1 is_number_char
+let is_leading_number_char = function | '0'..'9'|'-'|'+' -> true | _ -> false
+let leading_number_char =
+  peek_char >>=
+    (function
+     | None  -> fail "none"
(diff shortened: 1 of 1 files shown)


anmonteiro:
Perhaps #65 has a fix for this? Are there plans to merge the parser rewritten in Menhir?

andreas:
Thanks for reporting this. It seems like a test case is required 😄

Regarding Angstrom vs Menhir, I rewrote the parser in Menhir (#65) for comparison sake and to get a feel for the error reporting, which should be better out of the box (e.g. line/character numbers). Error reporting is indeed better, but it seems like you need to create and maintain a `.messages`-file to really get a big benefit.
On the other hand Angstrom supports parsing directly from a Bigarray, which is a win performance-wise, in combination with `httpaf`.

I originally started out trying to use `libgraphqlparser`, but I found the callback-style approach to be very awkward when trying to construct an AST in OCaml. There are also concerns regarding use with  Mirage.

Issue #42, opened 2017-04-30
How to query for other fields in a resolver before making http request?


sgrove:
I'm working with a REST backend that allows you to specify in the query params the fields you want along with some arguments about formatting. The problem right now is that when I enter that when I enter the top node, I have an `io_field` where I have to make the request, and then I pass on the result to the lower fields - but I actually want to know what those fields are so that I can make the appropriate request (rather than pulling out the very large full record all the time).

A few examples of a query and the url it would be translated into when resolving:

```
query {
  tweet(id:"123") {
    username
  }
}
```

`myapi.com/query?fields=username`

```
query {
  tweet(id:"123") {
    username
    body
  }
}
```

`myapi.com/query?fields=username,body`

```
query {
  tweet(id:"123") {
    username
    body(limit: 100)
  }
}
```

`myapi.com/query?fields=username,body[limit:100]`

In the twitter example size isn't such a big deal, but imagine lots of multi-megabyte fields in the db that I'd have to pull over by default right now when resolving `tweet` just in case the `reallyReallyBigField` was used, and also that there are *lots* of these kinds of fields so making different `tweetWithReallyReallyBigField { ... }` isn't really any option.

Any chance the resolver can be passed some context about its immediate children in a way that makes this easy to do?

andreas:
This seems like an entirely sane use case, and `graphql-js` also exposes a [`GraphqlResolveInfo`](https://github.com/graphql/graphql-js/blob/3c0a536688ae53df362c9777170e91aa66114760/src/type/definition.js#L598-L609)-object to resolvers:

```javascript
export type GraphQLResolveInfo = {
  fieldName: string;
  fieldNodes: Array<FieldNode>;
  returnType: GraphQLOutputType;
  parentType: GraphQLCompositeType;
  path: ResponsePath;
  schema: GraphQLSchema;
  fragments: { [fragmentName: string]: FragmentDefinitionNode };
  rootValue: mixed;
  operation: OperationDefinitionNode;
  variableValues: { [variableName: string]: mixed };
};
```

Is it something like that you had in mind?

The simplest solution would be to add an argument like this to all resolve functions. On the other hand I don't really like forcing all all resolve functions to receive this argument, when I expect the vast majority will not use it.

A solution which addresses this would be to introduce a separate function, say `field'`, for fields where the resolve function expects a resolve info argument:

```ocaml
val field' : ?doc:string ->
             string ->
             typ:('ctx, 'a) typ ->
[...]

PR #101, opened 2018-08-01
Subscription support

anmonteiro:
This is a WIP PR for adding subscription support to OCaml-GraphQL-Server.

### Things that are effectively implemented:

- [x] introspection for subscriptions
- [x] simple validation (according to the spec) that asserts a subscription only has one field 
- [x] calling the `subscribe` function

### Things that are still WIP (& for which feedback is greatly appreciated!)

- [x] settle on the type definitions. 
  - Right now a `SubscriptionField` has a `subscribe : 'ctx -> 'src -> 'args` and a `resolve : 'ctx -> 'out -> 'out` function. According to the spec, `resolve` is meant to be a mapper from `'out -> 'out`, and we want to return an `Io_stream.t` from `subscribe`. If we make both functions support `'args` I don't see a way to make them return different types. Maybe this is fine and `resolve` doesn't need args, but it'd be nice to have?
- [x] Execute the query in response to an event
  - Right now OGS gets the "source stream" from the `subscribe` function but not doing anything with it. This is the area where I have the most questions
  - The spec mentions that the "response stream" is basically a "mapper" over the source stream that executes the selection set of the subscription in response to any events in the stream. I think we have 2 alternatives here:
    1. Create a "mapped" stream from the source stream to which OGS pushes responses. This is effectively the "response stream" that the spec talks about, but we'd need a way to hand this response stream to the user (could be the result of executing but again the types don't match with the query execution types for example)
    2. Let the user do the job of setting up the stream and provide a helper to execute the selection set that the user can use when setting up the subscription behavior (this would need more info in the resolver – basically solving https://github.com/andreas/ocaml-graphql-server/issues/42); I'm leaning more towards this approach as it also offloads any decisions as to which types / libraries to use for streaming subscriptions to consumers of OGS – it's the less opinionated
  - another open question is if we want to support filtering events on the stream, which e.g. Apollo makes possible.

### DIFF
--- a/graphql-async/src/graphql_async.ml
+++ b/graphql-async/src/graphql_async.ml
@@ -1,5 +1,15 @@
-module Schema = Graphql_schema.Make(struct
+module Io = struct
   include Async_kernel.Deferred
 
   let bind x f = bind x ~f
-end)
+end
+
+module Stream = struct
+  type +'a io = 'a Async_kernel.Deferred.t
+  type 'a t = 'a Async_kernel.Pipe.Reader.t
+
+  let map x f = Async_kernel.Pipe.map' x ~f:(fun q ->
+    Async_kernel.Deferred.Queue.map q ~f)
+end
+
+module Schema = Graphql_schema.Make (Io) (Stream)
--- a/graphql-async/test/async_test.ml
+++ b/graphql-async/test/async_test.ml
@@ -16,12 +16,14 @@ let test_query schema ctx query expected =
     match Graphql_parser.parse query with
     | Error err -> failwith err
     | Ok doc ->
-      Graphql_async.Schema.execute schema ctx doc >>| fun result ->
-      let result' = match result with
-      | Ok data -> data
-      | Error err -> err
-      in
-      Alcotest.check yojson "invalid execution result" expected result'
+      Graphql_async.Schema.execute schema ctx doc >>= (function
+      | Ok (`Response data) -> Async_kernel.return data
+      | Ok (`Stream stream) ->
+          Async_kernel.Pipe.to_list stream >>| fun lst ->
+            `List Core_kernel.(List.map lst ~f:(fun x -> Option.value_exn (Result.ok x)))
+      | Error err -> Async_kernel.return err)
+      >>| fun result ->
+        Alcotest.check yojson "invalid execution result" expected result
   end
 
 let schema = Graphql_async.Schema.(schema [
@@ -35,6 +37,13 @@ let schema = Graphql_async.Schema.(schema [
         ~args:Arg.[]
         ~resolve:(fun () () -> Deferred.return (Ok 42))
     ]
+  ~subscriptions:[
+    subscription_field "int_stream"
+      ~typ:(non_null int)
+      ~args:Arg.[]
+      ~resolve:(fun () ->
+        Async_kernel.Deferred.Result.return (Async_kernel.Pipe.of_list [1; 2; 3]))
+  ]
 )
 
 let suite = [
@@ -46,6 +55,25 @@ let suite = [
       ]
     ])
   );
+  ("subscription", `Quick, fun () ->
+    test_query schema () "subscription { int_stream }" (`List [
+      `Assoc [
+        "data", `Assoc [
+          "int_stream", `Int 1
+        ]
+      ];
+      `Assoc [
+        "data", `Assoc [
+          "int_stream", `Int 2
+        ]
+      ];
+      `Assoc [
+        "data", `Assoc [
+          "int_stream", `Int 3
+        ]
+      ]
+    ])
+  )
 ]
 
 let () = Alcotest.run "graphql-server" ["async", suite]
--- a/graphql-lwt/src/graphql_lwt.ml
+++ b/graphql-lwt/src/graphql_lwt.ml
@@ -1,6 +1,16 @@
 open Graphql
 
-module Schema = Graphql_schema.Make(Lwt)
+module Io = Lwt
+
+module Stream = struct
+  type +'a io = 'a Lwt.t
+  type 'a t = 'a Lwt_stream.t * (unit -> unit)
+
+  let map (stream, destroy) f =
+    Lwt_stream.map_s f stream, destroy
+end
+
+module Schema = Graphql_schema.Make (Io) (Stream)
 
 module Server = struct
   module C = Cohttp_lwt_unix
@@ -33,7 +43,7 @@ module Server = struct
     Lwt_io.printf "Query: %s\n" query;
     let result = execute_query ctx schema (variables :> (string * Graphql_parser.const_value) list) operation_name query in
     result >>= function
-    | Ok data ->
+    | Ok (`Response data) ->
         let body = Yojson.Basic.to_string data in
         C.Server.respond_string ~status:`OK ~body ()
     | Error err ->
--- a/graphql-lwt/src/graphql_lwt.mli
+++ b/graphql-lwt/src/graphql_lwt.mli
@@ -1,6 +1,7 @@
 (** GraphQL schema with Lwt support *)
 module Schema : sig
   include Graphql_intf.Schema with type 'a io = 'a Lwt.t
+                              and type 'a stream = 'a Lwt_stream.t * (unit -> unit)
 end
 
 module Server : sig
--- a/graphql-lwt/test/lwt_test.ml
+++ b/graphql-lwt/test/lwt_test.ml
@@ -15,12 +15,17 @@ let test_query schema ctx query expected =
     match Graphql_parser.parse query with
     | Error err -> Lwt.fail_with err
     | Ok doc ->
-      Graphql_lwt.Schema.execute schema ctx doc >|= fun result ->
-      let result' = match result with
-      | Ok data -> data
-      | Error err -> err
-      in
-      Alcotest.check yojson "invalid execution result" expected result'
+      Graphql_lwt.Schema.execute schema ctx doc >>= (function
+      | Ok (`Response data) -> Lwt.return data
+      | Ok (`Stream (stream, destroy)) ->
+          Lwt_stream.to_list stream >|= fun lst ->
+            `List (
+              List.fold_right (fun x acc -> match x with
+                | Ok data -> data :: acc
+                | _ -> acc) lst [])
+      | Error err -> Lwt.return err)
+      >|= fun result ->
+        Alcotest.check yojson "invalid execution result" expected result
   end
 
 let schema = Graphql_lwt.Schema.(schema [
@@ -34,6 +39,15 @@ let schema = Graphql_lwt.Schema.(schema [
         ~args:Arg.[]
         ~resolve:(fun () () -> Lwt.return (Ok 42))
     ]
+  ~subscriptions:[
+    subscription_field "int_stream"
+      ~typ:(non_null int)
+      ~args:Arg.[]
+      ~resolve:(fun () ->
+        let stream = Lwt_stream.of_list [1; 2; 3] in
+        let destroy = (fun () -> ()) in
+        Lwt_result.return (stream, destroy))
+  ]
 )
 
 let suite = [
@@ -45,6 +59,25 @@ let suite = [
       ]
     ])
   );
+  ("subscription", `Quick, fun () ->
+    test_query schema () "subscription { int_stream }" (`List [
+      `Assoc [
+        "data", `Assoc [
+          "int_stream", `Int 1
+        ]
+      ];
+      `Assoc [
+        "data", `Assoc [
+          "int_stream", `Int 2
+        ]
+      ];
+      `Assoc [
+        "data", `Assoc [
+          "int_stream", `Int 3
+        ]
+      ]
+    ])
+  )
 ]
 
 let () = Alcotest.run "graphql-server" ["lwt", suite]
--- a/graphql/src/graphql.ml
+++ b/graphql/src/graphql.ml
@@ -1,6 +1,15 @@
-module Schema = Graphql_schema.Make(struct
+module Io = struct
   type +'a t = 'a
 
   let bind x f = f x
   let return x = x
-end)
+end
+
+module Stream = struct
+  type +'a io = 'a Io.t
+  type 'a t = 'a Seq.t
+
+  let map x f = Seq.map f x
+end
+
+module Schema = Graphql_schema.Make (Io) (Stream)
--- a/graphql/src/graphql_schema.ml
+++ b/graphql/src/graphql_schema.ml
@@ -33,9 +33,18 @@ module type IO = sig
   val bind : 'a t -> ('a -> 'b t) -> 'b t
 end
 
+(* Stream *)
+module type Stream = sig
+  type +'a io
+  type 'a t
+
+  val map : 'a t -> ('a -> 'b io) -> 'b t
+end
+
 (* Schema *)
-module Make(Io : IO) = struct
+module Make (Io : IO) (Stream: Stream with type 'a io = 'a Io.t) = struct
   type +'a io = 'a Io.t
+  type 'a stream = 'a Stream.t
 
   module Io = struct
     include Io
@@ -231,7 +240,7 @@ module Make(Io : IO) = struct
               eval_arglist variable_map arglist' key_values (f coerced)
             with StringMap.Missing_key key -> Error (Format.sprintf "Missing variable `%s`" key)
 
-    and eval_arg : type a. variable_map ->  a arg_typ -> Graphql_parser.const_value option -> (a, string) result = fun variable_map typ value ->
+    and eval_arg : type a. variable_map -> a arg_typ -> Graphql_parser.const_value option -> (a, string) result = fun variable_map typ value ->
       match (typ, value) with
       | NonNullable _, None -> Error "Missing required argument"
       | NonNullable _, Some `Null -> Error "Missing required argument"
@@ -329,14 +338,36 @@ module Make(Io : IO) = struct
   and ('ctx, 'a) abstract_value =
     AbstractValue : ('ctx, 'src option) typ * 'src -> ('ctx, 'a) abstract_value
 
+  type 'ctx subscription_field =
+    SubscriptionField : {
+      name       : string;
+      doc        : string option;
+      deprecated : deprecated;
+      typ        : ('ctx, 'out) typ;
+      args       : (('out stream, string) result io, 'args) Arg.arg_list;
+      resolve    : 'ctx -> 'args;
+    } -> 'ctx subscription_field
+
+  type 'ctx subscription_obj = {
+    name   : string;
+    doc    : string option;
+    fields : 'ctx subscription_field list;
+  }
+
   type ('ctx, 'a) abstract_typ = ('ctx, ('ctx, 'a) abstract_value option) typ
 
   type 'ctx schema = {
     query : ('ctx, unit) obj;
     mutation : ('ctx, unit) obj option;
+    subscription : 'ctx subscription_obj option;
   }
 
-  let schema ?(mutation_name="mutation") ?mutations ?(query_name="query") fields = {
+  let schema ?(mutation_name="mutation")
+             ?mutations
+             ?(subscription_name="subscription")
+             ?subscriptions
+             ?(query_name="query")
+             fields = {
     query = {
       name = query_name;
       doc = None;
@@ -349,6 +380,13 @@ module Make(Io : IO) = struct
         doc = None;
         abstracts = ref [];
         fields = lazy fields;
+      }
+    );
+    subscription = Option.map subscriptions ~f:(fun fields ->
+      {
+        name = subscription_name;
+        doc = None;
+        fields;
       }
     )
   }
@@ -366,6 +404,9 @@ module Make(Io : IO) = struct
 
   let abstract_field ?doc ?(deprecated=NotDeprecated) name ~typ ~args =
     AbstractField (Field { lift = Io.ok; name; doc; deprecated; typ; args; resolve = Obj.magic () })
+
+  let subscription_field ?doc ?(deprecated=NotDeprecated) name ~typ ~args ~resolve =
+    SubscriptionField { name; doc; deprecated; typ; args; resolve }
 
   let enum ?doc name ~values =
     Enum { name; doc; values }
@@ -395,6 +436,14 @@ module Make(Io : IO) = struct
         fun src -> AbstractValue (typ, src)
     | _ ->
         invalid_arg "Arguments must be Interface/Union and Object"
+
+  let obj_of_subscription_obj {name; doc; fields} =
+    let fields = List.map
+      (fun (SubscriptionField {name; doc; deprecated; typ; args; resolve}) ->
+        Field { lift = Obj.magic (); name; doc; deprecated; typ; args; resolve = (fun ctx () -> resolve ctx) })
+      fields
+    in
+    { name; doc; abstracts = ref []; fields = lazy fields }
 
   (* Built-in scalars *)
   let int : 'ctx. ('ctx, int option) typ = Scalar {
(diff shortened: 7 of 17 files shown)


anmonteiro:
I added a new commit to this PR that basically implements a hybrid between the 2 approaches I highlighted in the description above.

I added an example of how subscriptions would work with this PR in `examples/server.ml`. I feel like it's a nice and clean API (a pending question I have is in what order you would want the `source_stream` argument, i.e. before `src` or after).

I also solved the problem of passing arguments to the `resolve` function by parameterizing `Arg.arg_list` on more type variables. This, however, leads to a bit of duplication, so I'm open to alternatives if you see them.

Overall I think this is a pretty good approach, but I'm definitely open to any other thoughts you may have. There are still a few `TODO`s left in the code that are pretty easy to tackle, but would need to before we merge.

rrosenzweig:
Thanks for getting the ball rolling with subscriptions! I've tried to gather my thoughts below on the current approach and alternative approaches.

To clarify each of the approaches, I'll try use the "newMessage" example from the [GraphQL spec](http://facebook.github.io/graphql/June2018/#sec-Subscription). The following types are assumed in the examples below.

```ocaml
module type Schema = sig
  type +'a io
  type 'a stream

  type execute_error = [`Argument_error of string | ... ]
  type 'a response = ('a, execute_error) result

  (* Executing a GraphQL query will return either:
     - a single response
     - a stream of responses in case of subscriptions.
  *)
  val execute : ... -> [ `Single of Yojson.Basic.json
                       | `Stream of Yojson.Basic.json response stream
                       ] response io
end

type message = (* ... *)

module type PubSub = sig
  (* returns a stream of message ids *)
  val subscribe_to_new_messages : room_id:int -> int Lwt_stream.t
end

module type MessageStore = sig
  (* fetch a single message from database or similar *)
  val get : message_id:int -> (message, string) result Lwt.t
end
```

### Current Approach

This is an attempt to characterise the current approach:

- `resolve` maps a single event from the source stream to the output type: `'ctx -> 'src -> ('out, string) result io`, where `'src`. <br/>*Note:* The current draft implementation appears to be `'out -> 'out`, but it should be minor to change.
- `subscribe` receives an argument `'src stream -> Yojson.Basic.json response stream` and returns a stream of responses.<br/>*Note:* This part is not fully developed in the current draft implementation, but this is what I assume the intent is.

It looks roughly like this for the "newMessage" example:

```ocaml
Schema.(schema
  ~subscriptions:[
    subscription_field "newMessage"
      ~typ:(non_null message)
      ~args:Arg.[arg "roomId" ~typ:(non_null int)]
      ~resolve:(fun ctx message_id room_id ->
        MessageStore.get ~message_id
      )
      ~subscribe:(fun ctx source_stream_to_response_stream room_id ->
        (* source_stream_to_response_stream : message_id Lwt_stream.t -> Yojson.Basic.json Schema.response Lwt_stream.t *)
        let message_stream = PubSub.subscribe_to_new_messages ~room_id in
        source_stream_to_response_stream message_stream
      )
  ]
)
```

### Alternative 1

Simple twist of the current approach:

- `resolve` keeps the type from the current approach: `'src -> ('out, string) result io`.
- `subscribe` returns a source stream of type `'src stream` rather than a response stream.

The main benefit is a simplification from a user perspective, and the library can ensure that the application cannot return a response stream which contains invalid GraphQL responses (i.e. never calls `source_stream_to_response_stream`).

Example:

```ocaml
Schema.(schema
  ~subscriptions:[
    subscription_field "newMessage"
      ~typ:(non_null message)
      ~args:Arg.[arg "roomId" ~typ:(non_null int)]
      ~resolve:(fun ctx message_id room_id ->
        MessageStore.get ~message_id
      )
      ~subscribe:(fun ctx room_id ->
        PubSub.subscribe_to_new_messages ~room_id
      )
  ]
)
```

### Alternative 2

It's not clear to me why there's a distinction between `resolve` and `subscribe`. The closest I've come is [this a vague comment in `graphql/graphql-js`](https://github.com/graphql/graphql-js/pull/846#issuecomment-300388677), which talks about separating creation of a subscription and later executing in response to events. I would be interested in understanding this argument further.

Combining `resolve` and `subscribe` into a single function has some nice benefits in terms of simplicity though. In this case, `resolve` simply returns `('out, string) result stream`. Though my Scala-fu is not great, this also appears to be the approach taken by [`sangria`](https://github.com/sangria-graphql/sangria).

Example:

```ocaml
Schema.(schema
  ~subscriptions:[
    subscription_field "newMessage"
      ~typ:(non_null message)
      ~args:Arg.[arg "roomId" ~typ:(non_null int)]
      ~resolve:(fun ctx room_id ->
        PubSub.subscribe_to_new_messages room_id in
        |> Lwt_stream.map_s MessageStore.get
      )
  ]
)
```

anmonteiro:
I think we have different interpretations of the spec. I'll try my best below to address your comments and clarify the points which I think we view differently.

1. First off, you mention that we'd have to change the return type of the `execute` function. I did prototype something like that but in the end I do not think it's necessary. According to the spec, "Executing a subscription creates a persistent function on the server that maps an underlying Source Stream to a returned Response Stream". I think we can keep the return type of `(Yojson.Basic.json, Yojson.Basic.json) result io)` to signal that the creation of the persistent function was either successful or not.

The benefit of this is that the code remains simple as it currently is. Given that subscriptions are "not special" and they're a natural extension of queries and mutations, it _feels_ natural to me that the return type need not be special either.

2. Some of the things that you qualified as either under-developed or mistakenly so are in fact intentional, according to my understanding of the spec. But this is again dependent on how we interpret the spec, which I believe is intentionally vague. Here's how I think about the points you made:

> subscribe receives an argument 'src stream -> Yojson.Basic.json response stream and returns a stream of responses.
Note: This part is not fully developed in the current draft implementation, but this is what I assume the intent is.

I think this is very close to being complete, the only final piece being returning properly formatted JSON. It currently will return:

```json
{
  "sender": "Hagrid",
  "text": "You're a wizard!"
}
```

when we want it to return:

```json
{
  "data": {
    "newMessage": {
      "sender": "Hagrid",
      "text": "You're a wizard!"
    }
  }
}
```

> resolve maps a single event from the source stream to the output type: 'ctx -> 'src -> ('out, string) result io, where 'src. 
Note: The current draft implementation appears to be 'out -> 'out, but it should be minor to change.

The fact that the current draft implementation is `'out -> 'out` is intentional. According to the resources I've been studying, `resolve` in subscriptions is an opportunity to transform an event in the stream into another event of the _same_ type, in which the GraphQL execution engine will then extract the selection set. This is what the [Apollo docs](https://www.apollographql.com/docs/graphql-subscriptions/subscriptions-to-schema.html#subscription-server) say as well, which is where I got the inspiration from.

Furthermore, even though it's called the "source stream", my interpretation from the examples I've seen is that every event in the source stream is actually of the type `'out`. This is, I think, where our interpretations may differ the most, and the justification as to why I architected the current implementation as it is.

anmonteiro:
Following up on my response, I thought I'd give an example of what I'm talking about. Building on the `newMessage` example:

```ocaml
Schema.(schema
  ~subscriptions:[
    subscription_field "newMessage"
      ~typ:(non_null message)
      ~args:Arg.[arg "roomId" ~typ:(non_null int)]
      ~resolve:(fun ctx (message : message) room_id ->
        (* do something with message before passing it to the execution engine *)
        Message.censor_sensitive_language message
      )
      ~subscribe:(fun ctx source_stream_to_response_stream room_id ->
        (* source_stream_to_response_stream : message Lwt_stream.t -> Yojson.Basic.json result Lwt_stream.t *)
        let message_stream = PubSub.subscribe_to_new_messages ~room_id in
        source_stream_to_response_stream message_stream
      )
  ]
)
```

And for more information, here's another [example](https://github.com/notrab/graphql-server-tutorial/blob/master/src/mutations/createEvent.js#L21-L23) where the type of messages being published in the stream are of type `'out`.

anmonteiro:
I'm a little bit more convinced of my approach after seeing the reference implementation. This might be a little terse, but I think I now understand what [this code](https://github.com/graphql/graphql-js/blob/master/src/subscription/subscribe.js) is doing and it maps almost exactly to my approach.

anmonteiro:
I pushed yet another commit that returns the stream to the caller of `execute`. This is contrary to what I suggested above, but I read through the spec again and it [clearly says](http://facebook.github.io/graphql/June2018/#Subscribe()) the response stream should be returned as the result of execute, so let's do it.

The variant type looks like:

```ocaml
type execute_operation =
  [ `Single of Yojson.Basic.json
  | `Stream of (Yojson.Basic.json, Yojson.Basic.json) result Io_stream.t]
```

The variant tag names are as per your suggestion, but I think I would prefer `Json` and `Stream` (instead of `Single`).

Let me know what you think.

anmonteiro:
I pushed a commit addressing your latest comments, which really simplified the code.

I think this is ready for another round of review.

graphql/src/graphql_schema.ml
```diff
@@ -1089,28 +1161,97 @@ end
...
+  =
+    fun ctx src (SubscriptionField subs_field) field ->
+      let open Io.Infix in
+      let subscriber = subs_field.subscribe ctx.ctx src in
+      begin match Arg.eval_arglist ctx.variables subs_field.args field.arguments subscriber with
+      | Ok source_stream -> Io.ok (
+          Io_stream.map_s (fun value ->
+              present ctx value field subs_field.typ
+              |> Io.Result.map ~f:(fun (data, errors) ->
+                (data_to_json (`Assoc [field.name, data], errors)))
```
rrosenzweig: Should this be `alias_or_name field`?

graphql/src/graphql_schema.ml
```diff
@@ -328,15 +337,36 @@ module Make(Io : IO) = struct
...
   and ('ctx, 'a) abstract_value =
     AbstractValue : ('ctx, 'src option) typ * 'src -> ('ctx, 'a) abstract_value
+  and (_, _) subscription_field =
+    SubscriptionField : {
+      name       : string;
+      doc        : string option;
+      deprecated : deprecated;
+      typ        : ('ctx, 'out) typ;
+      args       : ('out io_stream, 'args) Arg.arg_list;
+      subscribe  : 'ctx -> 'src -> 'args;
```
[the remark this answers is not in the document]
anmonteiro: Your call here – let me know your decision

graphql/src/graphql_schema.ml
anmonteiro: Definitely. When I read [this](http://facebook.github.io/graphql/June2018/#sec-Source-Stream) section of the spec, I interpreted `Note: This value is unaffected if an alias is used.` as not returning an alias, but that's not what they mean. That section is only about validation and finding the subscription field.

graphql/src/graphql_schema.ml
```diff
@@ -33,9 +33,18 @@ module type IO = sig
   val bind : 'a t -> ('a -> 'b t) -> 'b t
 end
 
+(* IO Stream *)
+module type IO_Stream = sig
+  type +'a io
+  type 'a t
+
+  val map_s : ('a -> 'b io) -> 'a t -> 'b t
```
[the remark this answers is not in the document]
anmonteiro: I can do that. It might feel awkward to call it `map` and then use Lwt's `map_s` for the implementation though. What do you think?

graphql/src/graphql.ml
```diff
@@ -1,6 +1,16 @@
...
 
   let bind x f = f x
   let return x = x
-end)
\ No newline at end of file
+end
+
+module Io_stream = struct
+  type +'a io = 'a Io.t
+  type 'a t = 'a
```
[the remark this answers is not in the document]
anmonteiro: Went ahead and made it a list.

graphql/src/graphql_schema.ml
```diff
@@ -396,6 +437,14 @@ module Make(Io : IO) = struct
     | _ ->
         invalid_arg "Arguments must be Interface/Union and Object"
 
+  let obj_of_subscription_obj ({name; doc; abstracts; fields}: ('ctx, 'src) subscription_obj) =
+    let fields = List.map
+      (fun (SubscriptionField { name;doc; deprecated; typ; args; _}) ->
+        Field { lift = Obj.magic (); name; doc; deprecated; typ; args = Obj.magic args; resolve = Obj.magic () })
```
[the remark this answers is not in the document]
anmonteiro: Actually this was an artifact of having `'args` take 4 parameters; good catch – we don't need `Obj.magic` in this call at all.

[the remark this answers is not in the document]
anmonteiro: And we don't need the `Obj.magic` call either for `resolve`. It can just be `resolve = subscribe`

anmonteiro:
@rrosenzweig I pushed new commits that address your review comments.

graphql/src/graphql_schema.ml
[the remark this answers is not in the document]
rrosenzweig: I don't mind, I'm not so fond of `Lwt`'s naming conventions.

graphql/src/graphql_schema.ml
```diff
@@ -328,15 +337,36 @@ module Make(Io : IO) = struct
...
     AbstractValue : ('ctx, 'src option) typ * 'src -> ('ctx, 'a) abstract_value
+  and (_, _) subscription_field =
+    SubscriptionField : {
+      name       : string;
+      doc        : string option;
+      deprecated : deprecated;
+      typ        : ('ctx, 'out) typ;
+      args       : ('out io_stream, 'args) Arg.arg_list;
+      subscribe  : 'ctx -> 'src -> 'args;
+    } -> ('ctx, 'src) subscription_field
```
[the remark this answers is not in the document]
rrosenzweig: Ah yes, that was a typo.

graphql/src/graphql_schema.ml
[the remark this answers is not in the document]
rrosenzweig: I'd prefer `resolve` for consistency.

graphql-async/src/graphql_async.ml
```diff
@@ -2,4 +2,16 @@ module Schema = Graphql_schema.Make(struct
...
+  type 'a t = 'a Async_kernel.Pipe.Reader.t
+
+  let map x f = Async_kernel.Pipe.map' x ~f:(fun q ->
+    let open Core_kernel in
+    let open Async_kernel.Deferred.Infix in
+    Queue.fold ~f:(fun newq itm ->
+      (f itm) >>= fun itm ->
+        newq >>| fun newq ->
+        Queue.enqueue newq itm; newq)
+    ~init:(Async_kernel.Deferred.return (Queue.create ())) q)
```
rrosenzweig: Could you use `Queue.map q ~f` rather than the fold here?

graphql-lwt/src/graphql_lwt.mli
```diff
@@ -1,7 +1,8 @@
 (** GraphQL schema with Lwt support *)
 module Schema : sig
   include Graphql_intf.Schema with type 'a io = 'a Lwt.t
-end
+                              and type 'a stream = 'a Lwt_stream.t
+  end
```
rrosenzweig: Too much indentation on this line.

graphql/src/graphql.ml
```diff
@@ -1,6 +1,15 @@
...
+end
+
+module Stream = struct
+  type +'a io = 'a Io.t
+  type 'a t = 'a list
+
+  let map x f = List.map f x
+end
+
+module Schema = Graphql_schema.Make (Io) (Stream)
```
rrosenzweig: I think there should be consistency in how the functor `Graphql_schema.Make` is applied for `graphql.ml` (named modules), `graphql_lwt.ml` and `graphql_async.ml` (unnamed modules).

graphql/src/graphql.ml
```diff
@@ -1,6 +1,15 @@
...
 
   let bind x f = f x
   let return x = x
-end)
\ No newline at end of file
+end
+
+module Stream = struct
+  type +'a io = 'a Io.t
+  type 'a t = 'a list
```
rrosenzweig: Maybe it's silly to spend too much timing getting this type right, as a pubsub stream doesn't make much too much sense in a synchronous setting. However, after thinking more about it, I think `list` is a poor fit (though I did suggest it myself to begin with -- sorry about that). After looking more into `Stream`, `Seq` and `Queue`, I think `Seq` is the best option. Even though `Stream` is used as the stream type in [`ocaml-redis`](https://github.com/0xffea/ocaml-redis/blob/9a5c0ba79fc714ce719061349fd4457d9f119828/src_sync/redis_sync.ml#L12), it seems like it may `Stream` [may be deprecated](https://caml.inria.fr/mantis/view.php?id=6711). Note that this will require adding a dependency on `seq` for OCaml versions before 4.07.0.

graphql/src/graphql_schema.ml
```diff
@@ -329,14 +338,36 @@ module Make(Io : IO) = struct
   and ('ctx, 'a) abstract_value =
     AbstractValue : ('ctx, 'src option) typ * 'src -> ('ctx, 'a) abstract_value
 
+  type 'ctx subscription_field =
+    SubscriptionField : {
+      name       : string;
+      doc        : string option;
+      deprecated : deprecated;
+      typ        : ('ctx, 'out) typ;
+      args       : (('out stream, string) result, 'args) Arg.arg_list;
```
rrosenzweig: I would make this `('out stream, string) result io` to allow async operations to create the subscription. This is required e.g. for `ocaml-redis`.

graphql/src/graphql_schema.ml
```diff
@@ -396,6 +437,14 @@ module Make(Io : IO) = struct
     | _ ->
         invalid_arg "Arguments must be Interface/Union and Object"
 
+  let obj_of_subscription_obj {name; doc; fields} =
+    let fields = List.map
+      (fun (SubscriptionField {name; doc; deprecated; typ; args; resolve}) ->
+        Field { lift = Obj.magic (); name; doc; deprecated; typ; args; resolve = (fun ctx () -> resolve ctx) })
+      fields
+    in
+    { name; doc; abstracts = ref []; fields = lazy fields }
```
rrosenzweig: I've been thinking about how to avoid this, but I don't have a solution ready yet. Maybe it'll have to wait for second iteration on subscriptions.

graphql/src/graphql_schema.ml
```diff
@@ -971,11 +1035,12 @@ end
     ctx       : 'ctx;
   }
 
-  let matches_type_condition type_condition obj =
-    obj.name = type_condition ||
-      List.exists (fun (abstract : abstract) -> abstract.name = type_condition) !(obj.abstracts)
+  let matches_type_condition: string -> ('ctx, 'src) obj -> bool =
```
rrosenzweig: How about `let matches_type_condition type_condition (obj : ('ctx, 'src) obj) = ...`?

graphql/src/graphql_schema.ml
```diff
@@ -1089,28 +1160,98 @@ end
...
+          | Ok source_stream -> Io.ok (
+              Stream.map source_stream (fun value ->
+                present ctx value field subs_field.typ
+                |> Io.Result.map ~f:(fun (data, errors) ->
+                  (data_to_json (`Assoc [(alias_or_name field), data], errors)))
+                >>| to_response)
+            )
+          | Error err -> Io.error (`Resolve_error err)
+          end
+      | Error err -> Io.error (`Argument_error err)
```
rrosenzweig: Maybe the matches could be combined like so?

```ocaml
      match Arg.eval_arglist ctx.variables subs_field.args field.arguments resolver with
      | Ok (Ok source_stream) ->
          (* ... *)
      | Ok (Error err) ->
          Io.error (`Resolve_error err)
      | Error err ->
          Io.error (`Argument_error err)
```

graphql/src/graphql.ml
anmonteiro: Yeah, a `Seq` sounds appropriate, but I don't know how to add conditional dependencies to either Dune or Opam to only build on targets less than 4.07. Little help?

graphql/src/graphql_schema.ml
anmonteiro: I'm in favor of adding a `subscription_io_field` like we have for regular fields. I think some programs will want to have streams of values as currently implemented, while others like OCaml-redis would need `'out stream io`s.

I'll add an implementation in my next commit for you to see what you think

graphql/src/graphql_schema.ml
anmonteiro: This doesn't feel like a huge problem to me right now.

anmonteiro:
I pushed a new commit that addresses all the outstanding review items with the exception of the `Seq` one. I'm not sure how to conditionally add that library for OCaml < 4.07 only.

graphql/src/graphql_schema.ml
```diff
@@ -980,11 +1048,11 @@ end
     ctx       : 'ctx;
   }
 
-  let matches_type_condition type_condition obj =
-    obj.name = type_condition ||
-      List.exists (fun (abstract : abstract) -> abstract.name = type_condition) !(obj.abstracts)
+  let matches_type_condition type_condition (obj : ('ctx, 'src) obj) =
+      obj.name = type_condition ||
+        List.exists (fun (abstract : abstract) -> abstract.name = type_condition) !(obj.abstracts)
```
rrosenzweig: Too much indentation on these two lines.

graphql-lwt/src/graphql_lwt.ml
```diff
@@ -4,9 +4,10 @@ module Io = Lwt
 
 module Stream = struct
   type 'a io = 'a Lwt.t
-  include Lwt_stream
+  type 'a t = 'a Lwt_stream.t * (unit -> unit)
```
rrosenzweig: Does the "close function" need to be `unit -> unit io`?

graphql-async/test/async_test.ml
```diff
@@ -16,14 +16,26 @@ let test_query schema ctx query expected =
...
+        Alcotest.check yojson "invalid execution result" expected result
   end
 
+let take n l =
+  let rec loop n l = match n, l with
+    | 0, _ | _, [] -> []
+    | _, x::xs -> x :: loop (n - 1) xs
+  in
+  loop n l
+
```
rrosenzweig: Is it worthwhile to include this compared to just returning the full list from the `int_stream` field?

graphql-lwt/src/graphql_lwt.ml
anmonteiro: That may make more sense.

graphql-lwt/test/lwt_test.ml
```diff
@@ -15,14 +15,26 @@ let test_query schema ctx query expected =
...
+      >|= fun result ->
+        Alcotest.check yojson "invalid execution result" expected result
   end
 
+let take n l =
+  let rec loop n l = match n, l with
+    | 0, _ | _, [] -> []
+    | _, x::xs -> x :: loop (n - 1) xs
+  in
+  loop n l
```
rrosenzweig: Like the Async tests, I would consider omitting this.

graphql-async/test/async_test.ml
```diff
@@ -16,14 +16,26 @@ let test_query schema ctx query expected =
...
-      in
-      Alcotest.check yojson "invalid execution result" expected result'
+      Graphql_async.Schema.execute schema ctx doc >>= (function
+      | Ok (`Response data) -> Async_kernel.return data
+      | Ok (`Stream stream) ->
+          Async_kernel.Pipe.to_list stream >>| fun lst ->
+            `List (
+              List.fold_right (fun x acc -> match x with
+                | Ok data -> data :: acc
+                | _ -> acc) lst [])
```
rrosenzweig: Suggestion:

```ocaml
Async_kernel.Pipe.to_list stream >>| fun lst ->
`List Core_kernel.(List.filter_map lst ~f:Result.ok)
```

rrosenzweig: ... though I don't know if it's good we're ignoring the errors in the tests?

PR #102, opened 2018-08-02
Fix type error

revskill10:


anmonteiro:
What's the motivation behind this change? There's no type error in the code you're trying to change.

As far as I understand, the part of the example is meant to showcase `io_field` and shouldn't be removed.

revskill10:
@anmonteiro I'm sorry, i got type error in my case. Don't know about the showcase there.

anmonteiro:
@revskill10 you probably want to investigate what the error is in your particular setup. I can guarantee that the code does compile correctly.

anmonteiro:
That's not right... Did you modify the example in some other way?

revskill10:
@anmonteiro No, you can look at my commit history to see. I just follow the guide to get example working. I'm investigating the type error in my case.

PR #105, opened 2018-08-06
Error on unexisting field

### LINKED ISSUE #100: OGS doesn't validate that queried fields exist in the schema
The following query:
```graphql
query foo {
  somethingThatDoesntExistInQuery {
    ...
  }
}
```

returns
```json
{
  "data": {
    "somethingThatDoesntExistInQuery": null
  }
}
```

when it should probably result in a validation error (according to http://facebook.github.io/graphql/June2018/#sec-Executable-Definitions).

This happens at any level in the query, e.g. the following also doesn't produce a validation error:

```graphql
query {
  somethingThatExists {
    thisDoesntExist
  }
}
```

giltho:
Resolves #100 

Had to add the `subscriptionType` field to the `__Schema` type because the introspection query used contains it, but I kept the behavior it used to have (always returning `null` since it didn't exist).
That will create a very small merge conflict when merging #101

### DIFF
--- a/graphql/src/graphql_schema.ml
+++ b/graphql/src/graphql_schema.ml
@@ -942,6 +942,15 @@ module Introspection = struct
         args = Arg.[];
         lift = Io.ok;
         resolve = fun _ s -> []
+      };
+      Field {
+        name = "subscriptionType";
+        doc = None;
+        deprecated = NotDeprecated;
+        typ = __type;
+        args = Arg.[];
+        lift = Io.ok;
+        resolve = fun _ s -> None
       }
     ]
   }
@@ -1014,7 +1023,7 @@ end
     | Serial -> Io.map_s ~memo:[]
     | Parallel -> Io.map_p
 
-  let rec present : type ctx src. ctx execution_context -> src -> Graphql_parser.field -> (ctx, src) typ -> (Yojson.Basic.json * string list, [`Argument_error of string | `Resolve_error of string]) result Io.t =
+  let rec present : type ctx src. ctx execution_context -> src -> Graphql_parser.field -> (ctx, src) typ -> (Yojson.Basic.json * string list, [`Validation_error of string | `Argument_error of string | `Resolve_error of string]) result Io.t =
     fun ctx src query_field typ ->
       match typ with
       | Scalar s -> coerce_or_null src (fun x -> Io.ok (s.coerce x, []))
@@ -1042,7 +1051,7 @@ end
             present ctx (Some src') query_field typ'
           )
 
-  and resolve_field : type ctx src. ctx execution_context -> src -> Graphql_parser.field -> (ctx, src) field -> ((string * Yojson.Basic.json) * string list, [`Argument_error of string | `Resolve_error of string]) result Io.t =
+  and resolve_field : type ctx src. ctx execution_context -> src -> Graphql_parser.field -> (ctx, src) field -> ((string * Yojson.Basic.json) * string list, [`Validation_error of string | `Argument_error of string | `Resolve_error of string]) result Io.t =
     fun ctx src query_field (Field field) ->
       let open Io.Infix in
       let name = alias_or_name query_field in
@@ -1057,7 +1066,8 @@ end
           lifted_value >>| (function
           | Ok (value, errors) ->
               Ok ((name, value), errors)
-          | Error (`Argument_error _) as err ->
+          | Error (`Argument_error _)
+          | Error (`Validation_error _) as err ->
               err
           | Error (`Resolve_error err) as error ->
               match field.typ with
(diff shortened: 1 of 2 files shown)


anmonteiro:
@giltho I’m pretty sure my PR in #101 adds introspection support for subscriptions. What’s the motivation behind this change?

giltho:
@anmonteiro yes yes it does, there will be a small merge conflict. But one of the test in the `schema` test suits verifies that the introspection query works correctly. However, with my pull request, since `subscriptionType` was not part of the the `__Schema` type, that test was failing (returning an error saying `"Cannot query 'subscriptionType' on type '__Schema'" which is the expected behavioyr when it is not defined.

That is why I defined it with a dummy behaviour until it is redefined in #101

graphql/src/graphql_schema.ml
```diff
@@ -1057,19 +1066,20 @@ end
           lifted_value >>| (function
           | Ok (value, errors) ->
               Ok ((name, value), errors)
-          | Error (`Argument_error _) as err ->
-              err
+          | Error (`Argument_error _) as err -> err
+          | Error (`Validation_error _) as err -> err
```
rrosenzweig: These two patterns can be coalesced. For consistency, I'd also suggest `err` on it's own line like the other cases.

graphql/src/graphql_schema.ml
```diff
@@ -1057,19 +1066,20 @@ end
...
           | Error (`Resolve_error err) as error ->
               match field.typ with
               | NonNullable _ ->
                   error
               | _ ->
                   Ok ((name, `Null), [err])
+
```
rrosenzweig: Superfluous blank line.

graphql/src/graphql_schema.ml
```diff
@@ -1057,19 +1066,20 @@ end
...
                   error
               | _ ->
                   Ok ((name, `Null), [err])
+                
           )
       | Error err ->
           Io.error (`Argument_error err)
 
-  and resolve_fields : type ctx src. ctx execution_context -> ?execution_order:execution_order -> src -> (ctx, src) obj -> Graphql_parser.field list -> (Yojson.Basic.json * string list, [`Argument_error of string | `Resolve_error of string]) result Io.t =
+  and resolve_fields : type ctx src. ctx execution_context -> ?execution_order:execution_order -> src -> (ctx, src) obj -> Graphql_parser.field list -> (Yojson.Basic.json * string list, [`Validation_error of string | `Argument_error of string | `Resolve_error of string ]) result Io.t =
```
rrosenzweig: There's an additional space before the right bracket.

graphql/src/graphql_schema.ml
```diff
@@ -1079,7 +1089,9 @@ end
           | Some field ->
               resolve_field ctx src query_field field
           | None ->
-              Io.ok ((alias_or_name query_field, `Null), [])
+              let error_message = Printf.sprintf "Cannot query field '%s' on type '%s'" query_field.name obj.name in
```
rrosenzweig: To align with the wording of the spec, I'd suggest `"Field '%s' is not defined on type '%s'"`.

graphql/test/schema_test.ml
```diff
@@ -106,6 +106,26 @@ let suite = [
...
+  ("unexisting fields should be caught", `Quick, fun () ->
+    let query = "{ users { id foo } }" in
+    test_query query (`Assoc [
+      "errors", `List [
+        `Assoc [
+          "message", `String "Cannot query field 'foo' on type 'user'"
+        ]
+      ]
+    ])
+  );
```
rrosenzweig: Naming-wise, I'd suggest:

- `"undefined field on query root"`
- `"undefined field on object type"`

graphql/src/graphql_schema.ml
rrosenzweig: This seems to work for me:

```ocaml
          | Error (`Argument_error _)
          | Error (`Validation_error _) as err ->
              err
```

giltho: Oh turns out I didn't know the syntax correctly
you don't have to add `as err` on both lines
I'll fix that now

graphql/src/graphql_schema.ml
```diff
@@ -1079,7 +1089,9 @@ end
           | Some field ->
               resolve_field ctx src query_field field
           | None ->
-              Io.ok ((alias_or_name query_field, `Null), [])
+              let error_message = Printf.sprintf "Field '%s' is not defined on type '%s'" query_field.name obj.name in
+              Io.error (`Validation_error error_message)
+
```
rrosenzweig: Superfluous blank line.

giltho:
Done 😄

rrosenzweig:
Thanks for your contribution @giltho! 👍

PR #116, opened 2018-08-20
Add a graphql-lwt-cohttp package and move the Server module to it

### LINKED ISSUE #93: Example server impl in graphql-lwt prevents using Lwt 4.0
As far as I understand, the only reason Cohttp is required in `graphql-lwt` is to implement the simple server that serves GraphiQL.

This prevents the usage of `graphql-lwt` with Lwt >= 4.0 because `cohttp-lwt-unix` required lwt < 4.0.

My suggestion would be to remove the example server to either the examples folder or a new package altogether. 

What do you think?

anmonteiro:
fixes #93


rrosenzweig:
Would it be possible to have a package `graphql-cohttp` instead which allows the user to construct either a Lwt or Async Cohttp GraphQL server? Just to curb the proliferation of packages (`graphql-async-cohttp` comes to mind).

anmonteiro:
Sounds good to me. I'll rework the PR to account for that.

andreas:
Superceded by #126

PR #117, opened 2018-08-22
Add directive support and mandatory @skip and @include

### LINKED ISSUE #23: Support mandatory directives: @skip and @include
According to the spec, compliant servers must implement the directives `@skip` and `@include`:

https://facebook.github.io/graphql/#sec-Type-System.Directives

anmonteiro:
fixes #23

### DIFF
--- a/graphql/src/graphql_schema.ml
+++ b/graphql/src/graphql_schema.ml
@@ -1229,24 +1229,24 @@ end
     obj.name = type_condition ||
       List.exists (fun (abstract : abstract) -> abstract.name = type_condition) !(obj.abstracts)
 
-  let rec should_include_field ctx (directives : Graphql_parser.directive list) =
+  let rec should_include_field ctx ~field_name (directives : Graphql_parser.directive list) =
     let open Rresult in
     match directives with
     | [] -> Ok true
     | { name = "skip"; arguments }::rest ->
           let Directive directive = skip_directive in
           let resolver = directive.resolve in
-          Arg.eval_arglist ctx.variables directive.args arguments resolver >>= fun skip ->
+          Arg.eval_arglist ctx.variables ~field_name directive.args arguments resolver >>= fun skip ->
             let include_field = not skip in
             if include_field then
-              should_include_field ctx rest
+              should_include_field ctx ~field_name rest
             else Ok include_field
     | { name = "include"; arguments }::rest ->
           let Directive directive = include_directive in
           let resolver = directive.resolve in
-          Arg.eval_arglist ctx.variables directive.args arguments resolver >>= fun includ ->
+          Arg.eval_arglist ctx.variables ~field_name directive.args arguments resolver >>= fun includ ->
             if includ then
-              should_include_field ctx rest
+              should_include_field ctx ~field_name rest
             else Ok includ
     | { name; _ }::_ ->
         let err = Format.sprintf "Unknown directive: %s" name in
@@ -1257,26 +1257,27 @@ end
     let open Rresult in
     List.map (function
     | Graphql_parser.Field field ->
-        should_include_field ctx field.directives >>| fun include_field ->
+        should_include_field ctx ~field_name:field.name field.directives >>| fun include_field ->
           if include_field then [field] else []
     | Graphql_parser.FragmentSpread spread ->
         begin match StringMap.find spread.name ctx.fragments with
-          | Some fragment when matches_type_condition fragment.type_condition obj ->
-            should_include_field ctx fragment.directives >>= fun include_field ->
-              if include_field then
-                collect_fields ctx obj fragment.selection_set
-              else Ok []
-          | _ -> Ok []
+        | Some { name; directives; type_condition; selection_set }
+          when matches_type_condition type_condition obj ->
+          should_include_field ctx ~field_name:name directives >>= fun include_field ->
+            if include_field then
+              collect_fields ctx obj selection_set
+            else Ok []
+        | _ -> Ok []
         end
     | Graphql_parser.InlineFragment fragment ->
         match fragment.type_condition with
         | None ->
-          should_include_field ctx fragment.directives >>= fun include_field ->
+          should_include_field ctx ~field_name:"inline fragment" fragment.directives >>= fun include_field ->
             if include_field then
               collect_fields ctx obj fragment.selection_set
             else Ok []
         | Some condition when matches_type_condition condition obj ->
-          should_include_field ctx fragment.directives >>= fun include_field ->
+          should_include_field ctx ~field_name:"inline fragment" fragment.directives >>= fun include_field ->
             if include_field then
               collect_fields ctx obj fragment.selection_set
             else Ok []
@@ -1463,7 +1464,7 @@ end
           )
       | Error err -> Io.error (`Argument_error err)
 
-  let execute_operation : 'ctx schema -> 'ctx execution_context -> Graphql_parser.operation -> ([ `Response of Yojson.Basic.json | `Stream of Yojson.Basic.json response stream], [> execute_error]) result Io.t =
+  let execute_operation : 'ctx schema -> 'ctx execution_context -> Graphql_parser.operation -> ([ `Response of Yojson.Basic.json | `Stream of Yojson.Basic.json response Io.Stream.t], [> execute_error]) result Io.t =
     fun schema ctx operation ->
       let open Io.Infix in
       match operation.optype with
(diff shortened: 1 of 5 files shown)


graphql/src/graphql_schema.ml
```diff
@@ -356,41 +360,33 @@ module Make (Io : IO) (Stream: Stream with type 'a io = 'a Io.t) = struct
...
+
+  type ('ctx, 'out) directive =
+    Directive : {
+      name       : string;
+      doc        : string option;
+      locations  : directive_location list;
+      typ        : ('ctx, 'out) typ;
+      args       : ('out, 'args) Arg.arg_list;
+      resolve    : 'ctx -> 'args;
+    } -> ('ctx, 'out) directive
```
rrosenzweig: To simplify, I would propose:

```ocaml
  type directive =
    Directive : {
      name       : string;
      doc        : string option;
      locations  : directive_location list;
      args       : (bool, 'args) Arg.arg_list;
      resolve    : 'args;
    } -> directive
```

graphql/src/graphql_schema.ml
```diff
@@ -437,14 +433,6 @@ module Make (Io : IO) (Stream: Stream with type 'a io = 'a Io.t) = struct
...
         invalid_arg "Arguments must be Interface/Union and Object"
 
-  let obj_of_subscription_obj {name; doc; fields} =
-    let fields = List.map
-      (fun (SubscriptionField {name; doc; deprecated; typ; args; resolve}) ->
-        Field { lift = Obj.magic (); name; doc; deprecated; typ; args; resolve = (fun ctx () -> resolve ctx) })
-      fields
-    in
-    { name; doc; abstracts = ref []; fields = lazy fields }
-
```
rrosenzweig: What are your thoughts on moving these lines (and the ones above) further down?

graphql/src/graphql_schema.ml
```diff
@@ -1048,26 +1179,65 @@ end
...
+      let resolver = include_directive.resolve () in
+      Arg.eval_arglist ctx.variables include_directive.args arguments resolver
+    | Some skip, Some includ ->
+        let skip_resolver = skip_directive.resolve () in
+        let skip_args = skip.arguments in
+        let include_resolver = include_directive.resolve () in
+        let includ_args = includ.arguments in
+        Arg.eval_arglist ctx.variables skip_directive.args skip_args skip_resolver >>= fun skip_res ->
+        Arg.eval_arglist ctx.variables include_directive.args includ_args include_resolver >>| fun include_res ->
+          not skip_res && include_res
```
rrosenzweig: I would propose a recursive definition along these lines:

```ocaml
  let rec should_include ctx (directives : Graphql_parser.directive list) =
    match directives with
    | [] -> Ok true
    | { name = "skip"; arguments }::rest ->
        (* ... *)
    | { name = "include"; arguments }::rest ->
        (* ... *)
    | { name; _ }::_ ->
        let err = Format.sprintf "Unknown directive: %s" name in
        Error err
```

graphql/src/graphql_schema.ml
```diff
@@ -1048,26 +1179,65 @@ end
...
+
+  let rec collect_fields : 'ctx execution_context -> ('ctx, 'src) obj -> Graphql_parser.selection list -> (Graphql_parser.field list, string) result =
+    fun ctx obj fields ->
+    let open Rresult in
     List.map (function
     | Graphql_parser.Field field ->
-        [field]
+        should_include ctx field.directives >>| fun should_include ->
```
rrosenzweig: I think it's unfortunate with the name collision.

graphql/src/graphql_schema.ml
anmonteiro: Where exactly? I've moved them just before the introspection module which feels ok to me.

anmonteiro:
Addressed your comments and rebased on master.

graphql/src/graphql_schema.ml
```diff
@@ -7,6 +7,10 @@ module List = struct
   let find_exn = find
   let find cond xs = try Some (find_exn cond xs) with Not_found -> None
 
+  let rec find_opt p = function
+    | [] -> None
+    | x :: l -> if p x then Some x else find_opt p l
+
```
rrosenzweig: This seems unused now.

graphql/src/graphql_schema.ml
```diff
@@ -356,41 +360,33 @@ module Make (Io : IO) (Stream: Stream with type 'a io = 'a Io.t) = struct
...
+    | `Inline_fragment
+    | `Variable_definition
+    ]
+
+  type directive =
+    Directive : {
+      name       : string;
+      doc        : string option;
+      locations  : directive_location list;
+      typ        : ('ctx, 'out) typ;
```
rrosenzweig: This field is no longer required, I think.

graphql/src/graphql_schema.ml
rrosenzweig: Sorry, I was being unclear: why move them at all in this PR? 😄

graphql/src/graphql_schema.ml
```diff
@@ -1068,26 +1199,54 @@ end
...
+          let Directive directive = skip_directive in
+          let resolver = directive.resolve in
+          Arg.eval_arglist ctx.variables directive.args arguments resolver
+          |> Rresult.R.map (fun skip -> not skip && should_include_rest)
+    | { name = "include"; arguments }::rest ->
+        should_include ctx rest >>= fun should_include_rest ->
+          let Directive directive = include_directive in
+          let resolver = directive.resolve in
+          Arg.eval_arglist ctx.variables directive.args arguments resolver
+          |> Rresult.R.map (fun includ -> includ && should_include_rest)
```
rrosenzweig: Instead of recursing first and considering the current element later, how considering the current element first? This allows you to short-circuit in case the current element indicates the field should not be included.

graphql/src/graphql_schema.ml
```diff
@@ -1068,26 +1199,54 @@ end
...
+        let err = Format.sprintf "Unknown directive: %s" name in
+        Error err
+
+  let rec collect_fields : 'ctx execution_context -> ('ctx, 'src) obj -> Graphql_parser.selection list -> (Graphql_parser.field list, string) result =
+    fun ctx obj fields ->
+    let open Rresult in
     List.map (function
     | Graphql_parser.Field field ->
-        [field]
+        should_include ctx field.directives >>| fun should_include_field ->
```
rrosenzweig: I would suggest giving the variable the short name and the function the longer name. Maybe the function `should_include_field` and the variable `include_field`?

graphql/src/graphql_schema.ml
```diff
@@ -476,6 +464,67 @@ module Make (Io : IO) (Stream: Stream with type 'a io = 'a Io.t) = struct
...
+  (* Mandatory directives: skip and include *)
+  let skip_directive = Directive {
+    name = "skip";
+    doc = Some "Directs the executor to skip this field or fragment when the `if` argument is true.";
+    locations = [`Field; `Fragment_spread; `Inline_fragment];
+    typ = non_null bool;
+    args = Arg.[
+      arg "if" ~doc:"Skipped when true." ~typ:(non_null bool)
+    ];
+    resolve = fun if_ -> if_
```
rrosenzweig: Maybe return `not if_` here to make the directives homogenous in the semantics of their return value (i.e. whether the field should be included or not)?

graphql/src/graphql_schema.ml
anmonteiro: Oh we actually don't need to. I was envisioning passing some directives argument to `schema`, which would mean their definition would have to come before.

graphql/src/graphql_schema.ml
anmonteiro: Interesting thought – I think that conflates the directive definition with its usage in `should_include`, though. 

I'm imagining that we could want the raw value of `"skip"` in some future usage (i.e. if the spec evolves).

anmonteiro:
Reworked the PR to include your latest comments.

anmonteiro:
I had forgotten about this PR until I watched @sgrove's talk at Øredev 2018 where he mentioned directives were still WIP 😅 

I just rebased this on master and fixed the tests. @andreas could you take another look at this please?

graphql/src/graphql_schema.ml
andreas: What if the return type was ```[`Skip | `Include]```?

anmonteiro:
@andreas addressed the latest review comments and rebased on master.

andreas:
Thanks a lot! 🏆

PR #125, opened 2018-10-07
Update instructions to run the example project

naartjie:
To make it work, I had to `opam install async_kernel`.

This was on a clean `opam switch create 4.07.0` and `opam switch create 4.07.1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ API documentation:
 To run a sample GraphQL server also serving GraphiQL, do the following:
 
 ```bash
-opam install graphql-lwt dune async_kernel
+opam install graphql-lwt dune seq
 git clone git@github.com:andreas/ocaml-graphql-server.git
 cd ocaml-graphql-server/examples
 dune exec ./server.exe
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ API documentation:
 To run a sample GraphQL server also serving GraphiQL, do the following:
 
 ```bash
-opam install graphql-lwt dune
+opam install graphql-lwt dune async_kernel
 git clone git@github.com:andreas/ocaml-graphql-server.git
 cd ocaml-graphql-server/examples
 dune exec ./server.exe


naartjie:
So I just tried again with `4.06.1`, running `dune exec ./server.exe` in `./examples`:

```
File "graphql/src/dune", line 5, characters 42-45:
5 |  (libraries graphql_parser yojson rresult seq)
                                              ^^^
Error: Library "seq" not found.
Hint: try: dune external-lib-deps --missing ./server.exe
```

If I run `make` first, from the project root:
```
dune build @install @DEFAULT
File "graphql-async/src/dune", line 6, characters 20-32:
6 |  (libraries graphql async_kernel))
                        ^^^^^^^^^^^^
Error: Library "async_kernel" not found.
Hint: try: dune external-lib-deps --missing @install @DEFAULT
File "graphql/src/dune", line 5, characters 42-45:
5 |  (libraries graphql_parser yojson rresult seq)
                                              ^^^
Error: Library "seq" not found.
Hint: try: dune external-lib-deps --missing @install @DEFAULT
ocaml-crunch graphql-lwt/src/assets.ml
Generating assets.ml
Skipping generation of .mli
make: *** [all] Error 1
```

If I run the recommended `dune external-lib-deps --missing @install @DEFAULT`, I then get:

```
Done: 0/0 (jobs: 0)Error: The following libraries are missing in the default context:
- async_kernel
- seq
Hint: try: opam install async_kernel seq
```

If I run `opam install async_kernel seq`, then `dune exec ./examples/server.exe` from project root starts the examples server fine.

naartjie:
Is it this line?

https://github.com/andreas/ocaml-graphql-server/blob/84fbed6544c82d7218a49e39783e8fb3974ee55f/graphql-async.opam#L13

anmonteiro:
@naartjie There are a couple things I still don't understand.

1. Have you run `opam install ocaml-graphql-server` like the instructions mention?

2. You're trying to run `make` but the instructions never mention that. Therefore, it's expected that you're getting the error you're getting, because `make` will build **all** targets, including `graphql-async` (which you're not going to use)

andreas:
I expect this is resolved now. Please open again otherwise.

naartjie:
I checked with a clean `opam switch create 4.07.1` and this is the full list of dependencies I needed to install to make the example work:

```
$ opam install graphql-lwt dune astring base64 cohttp cohttp-lwt-unix digestif ocplib-endian crunch
```

I know it might seem like nitpicking at this stage and any OCaml user would be able to immediately pick up on what is missing, but I think it might be helpful to a newcomer at least, and save some time.

I couldn't reopen this PR, so I created #157 in case you want to merge that in.

wokalski:
@naartjie are they not all included in the opam files? If they are it should “just work”

naartjie:
@wokalski there are opam files for the libraries, but not for `./examples/server.ml`. Maybe I'm missing something though...

Issue #121, opened 2018-10-01
Guidance for integrating with existing Opium/Other server. 


anisjonischkeit:
I have an existing Opium server and just want one of my endpoints to be graphql. If anyone can point me in the right direction for this I'd be really appreciative.

Also sorry in advance if this is the wrong place to ask.

anmonteiro:
Here's an example that you could copy: https://github.com/andreas/ocaml-graphql-server/blob/cad6b3c6bf542ac5d848631dc3846ea08be70c1b/graphql-lwt/src/graphql_lwt.ml#L31-L54

andreas:
There's no proper integration between `opium` and `ocaml-graphql-server` right now. Copying some of the code from `Graphql_lwt.Server` will be the quickest way to get going (what @anmonteiro mentioned).

There's previously been talk about exposing a function á la `'ctx Graphql_lwt.Schema.t -> 'ctx -> Cohttp.Request.t -> Cohttp.Body.t -> (Yojson.Basic.json, Yojson.Basic.json) result Lwt.t`, which seems like it would be helpful in this scenario as well, but it's not there yet.

anisjonischkeit:
Thanks for the guidance, it's really helped. Just thought I'd show what I came up with in-case someone else gets stuck with this.

```ocaml
open Opium.Std
open Graphql_lwt

type user = { 
  id : int;
  name : string;
} 
  
let alice = { id = 1; name = "Alice" }
and bob = { id = 2; name = "Bob"}
 
let users = [alice; bob]

let user = Schema.(obj "user"
  ~fields:(fun _user -> [
    field "id"
      ~args:Arg.[]
      ~typ:(non_null int)
      ~resolve:(fun () p -> p.id)
    ;
    field "name"
      ~args:Arg.[]
      ~typ:(non_null string)
      ~resolve:(fun () p -> p.name)
    ;
  ])
)

let schema = Schema.(schema [
    io_field "users"
      ~args:Arg.[]
      ~typ:(non_null (list (non_null user)))
      ~resolve:(fun () () -> Lwt_result.return users)
    ;
  ])


let json_err = function
  | Ok _ as ok -> ok
  | Error err -> Error (`String err)

let execute_query ctx schema variables operation_name query =
  let open Lwt_result in
  Lwt.return @@ json_err @@ Graphql_parser.parse query >>= fun doc ->
  Schema.execute schema ctx ~variables ?operation_name doc

[...]

PR #126, opened 2018-10-08
New package: graphql-cohttp

andreas:
This PR introduces a new package `graphql-cohttp` for serving GraphQL requests via HTTP using the `cohttp` library. It generalises the server previously found in `Graphql_lwt.Server` to be IO-agnostic, and thus should work with both `Async`, `Lwt` with Unix, and `Lwt` with Mirage.

The PR is an alternative approach compared to #116. However, rather than try to encompass running a HTTP server, which makes the required functor arguments very complex, it only provides two capabilities:

1) Executing a single HTTP request against a GraphQL schema.
2) Constructing a callback function to be used with a `Cohttp_async.Server`, `Cohttp_lwt_unix.Server` or `Cohttp_mirage.Server`.

This has the benefits that the functor argument is quite simple, and it exposes some primitives that compose better with other libraries (e.g. `opium`, see #121).

Running a server is slightly more code than with `Graphql_lwt.Server`, but I think the above benefits outweigh this drawback. Example:

```ocaml
module HttpBody = struct
  include Cohttp_lwt.Body

  type +'a io = 'a Lwt.t
  type 'a stream = 'a Lwt_stream.t * (unit -> unit)

  let of_stream (stream, _) = Cohttp_lwt.Body.of_stream stream
end

module Graphql_cohttp_lwt = Graphql_cohttp.Make (Graphql_lwt.Schema) (HttpBody)

let callback = Graphql_cohttp_lwt.make_callback (fun _req -> ()) schema in
let server = Cohttp_lwt_unix.Server.make ~callback () in
let mode = `TCP (`Port 8080) in
Cohttp_lwt_unix.Server.create ~mode server
```
(if `Lwt_stream.t` supported closing, this would be much simpler -- the `Async` case is very simple)

A couple of other changes have been lumped into this PR:

- The `Stream` module has been moved into `Io`, since `Stream` depends on `Io` anyway.
- `Graphql_intf.Schema` exposes the provided `Io` module.
- The HTTP server supports returning subscription results as Server-Sent Events.


zshipko:
What's the status of this? Are you planning on any further work before you consider this ready to be released?

I have been using this branch to create a version of `irmin-graphql` that supports both Unix and MirageOS ([link](https://github.com/zshipko/irmin/blob/graphql-changes-to-support-mirage/src/irmin-graphql/irmin_graphql.ml)) and `graphql-cohttp` has been working very well for me!

andreas:
My goal with `graphql-cohttp` was to create a fairly generic package that would work somewhat easily with `cohttp-async`, `cohttp-lwt-unix` and `cohttp-mirage`, such that I would avoid creating a package for each of those.

I think that's succeeded fairly well with the current implementation, though with one exception: streams created when issuing a subscription query are not destroyed when the connection is later closed. Unfortunately the handling of closed connections is quite different between `cohttp-async` and `cohttp-lwt`:

- `cohttp-async` exposes a `Reader.t` to the handler, and provides `Reader.closed : Reader.t -> unit Deferred.t` to know when the connection is closed.
- `cohttp-lwt` allows you to provide a "hook function" `conn_closed : (IO.conn, Cohttp.Connection.t) -> unit` to `Cohttp_lwt_unix.Server.make` and `Cohttp_mirage.Server.make`. As such, this is not at the callback-level, but the server-level.

This makes it complicated to maintain the homogenous interface whilst not requiring too much boilerplate from the user of the library. I also worry that the amount of boilerplate will increase when a websocket transport layer is introduced 😢 I'm exploring different options of changing `cohttp` to make this more feasible, but I'm not sure if they would accept such changes (e.g. moving `conn_closed` to the callback-level for `cohttp-lwt`).

zshipko:
Thanks for the explanation! I am not very familiar with cohttp-async but am surprised to hear about the interface incompatibilities.

It seems like the best bet for irmin-graphql, until these issues are solved, might be to create our own graphql server implementation which can support Unix and Mirage, since only Lwt is used.

andreas:
> It seems like the best bet for irmin-graphql, until these issues are solved, might be to create our own graphql server implementation which can support Unix and Mirage, since only Lwt is used.

That's a possibility, though I would really like to consolidate the HTTP transport code so everybody can benefit from improving one implementation.

I have an idea that will shift the burden of signaling closed connections onto the user of `graphql-cohttp`, but with an easy escape hatch if you don't use subscriptions:

```diff
  type 'conn callback =
    'conn ->
+    conn_closed:unit Schema.Io.t ->
    Cohttp.Request.t ->
    Body.t ->
    (Cohttp.Response.t * Body.t) Schema.Io.t

  val execute_request :
    'ctx Schema.schema ->
    'ctx ->
+    conn_closed:unit Schema.Io.t ->
    Cohttp.Request.t ->
    Body.t ->
    (Cohttp.Response.t * Body.t) Schema.Io.t

  val make_callback :
    (Cohttp.Request.t -> 'ctx) ->
    'ctx Schema.schema ->
    'conn callback
```

The promise `conn_closed` should be resolved when the connection is closed to allow tearing down subscription streams. If you don't use subscriptions, you can simply provide a promise that never resolves.

This functor adds the required additional boilerplate code to signal closing of connections properly for a given `Cohttp_lwt.S.Server`:

```ocaml
module ServerWithClose (S : Cohttp_lwt.S.Server) = struct
  include S

  let make ~callback () =
    let close_handlers = Hashtbl.create 8 in
    let conn_closed (conn_id, _conn) =
      try Hashtbl.find close_handlers conn_id ()
      with Not_found -> ()
    in
    let callback' (conn_id, conn) req body =
      let conn_closed, wakener = Lwt.wait () in
      Hashtbl.add close_handlers conn_id (Lwt.wakeup_later wakener);
      callback (conn_id, conn) ~conn_closed req body
    in S.make ~conn_closed ~callback:callback' ()
end
```

Long-term, this should handled by the library without requiring users to adding boilerplate code.

Do you think this could work for you?

zshipko:
Yeah, that would work. It is *slightly* inconvenient to have to handle this but definitely much better than re-implementing the server!

I’m happy to update my branch to test it out once these changes are made to `graphql-cohttp`.

andreas:
After more investigation, I've submitted the following PR against `cohttp`: https://github.com/mirage/ocaml-cohttp/pull/647

If that's merged, then adding websocket support can be done quite elegantly with support for both Async and Lwt -- and with minimal boilerplate code! As such, I've removed support for subscription responses using Server-Sent Events in this PR, since it's somewhat messy. The intent is then to follow up with another PR adding websocket support, which is arguably much more useful anyway.

zshipko:
Great! That approach seems much nicer :smile: 

I've already updated my code to work with your new changes and everything looks good. Definitely looking forward to the websocket-based subscription notifications!

PR #128, opened 2018-10-11
Improve error messages for invalid arguments

dwwoelfel:
I had a really hard time discovering the source of an error for a script that was making GraphQL queries to OneGraph. The error was "Invalid string", but I had multiple string arguments. 

With this change, the error I get back is ``Argument `pos` of type `String` expected on field `updateCard`, found 1.``

With that error, there is enough context for me to easily pinpoint the problem and fix it.

If you'd like an easy way to test, this change is live on OneGraph:

```
curl 'https://serve.onegraph.com/dynamic?app_id=0b33e830-7cde-4b90-ad7e-2a39c57c0e11'  \
    --data-binary '{"query":"{ trello { card { name }}}"}'

> {"data":null,"errors":[{"message":"Argument `id` of type `String!` expected on field `card`, but not provided."}]}
```


dwwoelfel:
Thanks @andreas! Updated per your comments.

Any chance we could use ocamlformat on this project? https://github.com/ocaml-ppx/ocamlformat

Issue #119, opened 2018-09-03
June 2018 Spec compliance checklist


anmonteiro:
The following is an incomplete but growing list of things that we need to support to achieve compliance with the Jun 2018 version of the Spec.

- [ ] [Block strings](http://facebook.github.io/graphql/June2018/#sec-String-Value) (using `"""`)
- [ ] Check that enum values are not `true`, `false` or `null` (http://facebook.github.io/graphql/June2018/#sec-Enum-Value)
- [ ] [Field Selection Merging](http://facebook.github.io/graphql/June2018/#sec-Field-Selection-Merging)
- [ ] Validate that [fragments must be used](http://facebook.github.io/graphql/June2018/#sec-Fragments-Must-Be-Used) if defined in a document
- [x] the `errors` key [must appear first](http://facebook.github.io/graphql/June2018/#sec-Response-Format) if present (#131)
- [ ] [Schema](http://facebook.github.io/graphql/June2018/#sec-Schema-Extension) and [object](http://facebook.github.io/graphql/June2018/#sec-Object-Extensions) extensions

PR #131, opened 2018-12-09
Add the `errors` key first in the response JSON

anmonteiro:
According to the `Response Format`[section](https://facebook.github.io/graphql/June2018/#sec-Response-Format) of the GraphQL Spec:

> When errors is present in the response, it may be helpful for it to
appear first when serialized to make it more clear when errors are
present in a response during debugging.

refs #119


andreas:
Thanks, nice improvement! 👍

PR #135, opened 2019-01-26
Fix parsing arguments with default values in positions other than last

anmonteiro:
cc @sgrove who reported this to me.

### DIFF
--- a/graphql_parser/src/graphql_parser.ml
+++ b/graphql_parser/src/graphql_parser.ml
@@ -277,7 +277,7 @@ let variable_definition = lift3' (fun name typ default_value -> {
   name;
   typ;
   default_value;
-}) (dollar *~> name <~* colon) typ (optional (equal *~> const_value))
+}) (dollar *~> name <~* colon) typ (optional (equal *~> const_value <* ignored))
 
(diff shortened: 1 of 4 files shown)


graphql_parser/src/graphql_parser.ml
```diff
@@ -277,7 +277,7 @@ let variable_definition = lift3' (fun name typ default_value -> {
   name;
   typ;
   default_value;
-}) (dollar *~> name <~* colon) typ (optional (equal *~> const_value))
+}) (dollar *~> name <~* colon) typ (optional (equal *~> const_value <* ignored))
```
anmonteiro: `lift3'` doesn't `<* ignored` the last argument.

andreas:
Thanks for pointing this out and coming up with a test case!

I've tried to follow the rule that "grammar elements" should consume ignored characters *before it*, but not after. This allows elements to compose nicely without backtracking on ignored characters. That's also why `lift3'` does not consume ignored characters after the last element.

As such, I'd prefer updating `variable_definitions` to use `sep_by ignored variable_definition` rather than have `variable_definition` consume ignored characters *after it*.

With that being said, I'm about to merge #132, which I've confirmed fixes this bug also.

PR #139, opened 2019-02-08
Minor parser improvements

keleshev:
Really cool to see the parser switch to Menhir. I thought I might chip in with a few minor parser improvements. Cherry-pick the ones you find useful.

### DIFF
--- a/graphql_parser/src/parser.mly
+++ b/graphql_parser/src/parser.mly
@@ -87,7 +87,7 @@ selection:
   | inline_fragment { $1 }
 
 field:
-  | name COLON name loption(arguments) directive* loption(selection_set)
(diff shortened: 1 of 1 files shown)


andreas:
Thanks, these look like great improvements! I prefer keeping it explicit that arguments are optional (eace7b1) though, even if it's not as terse. I'll cherry-pick and merge once v0.9.0 is out.

Don't mind the CI failure: I haven't found a good way to bump version constraints prior to getting the new package released on OPAM.

If you have any further interest in contributing to `graphql_parser`, I would love to rope you into adding support for "block strings" ([spec](https://facebook.github.io/graphql/June2018/#sec-String-Value)) 😉

andreas:
Cherry-picked and pushed to master. Thanks!

PR #140, opened 2019-02-09
Use Format for pretty-printing instead of Fmt

keleshev:
This is probably a bit random, but I replaced `Fmt` with the standard `Format` module. One less dependency, plus if you like `printf` you might find this more readable.

I hope to get onto something more useful (like block strings) soon

### DIFF
--- a/graphql_parser.opam
+++ b/graphql_parser.opam
@@ -17,7 +17,6 @@ depends: [
   "menhir" {build}
   "alcotest" {with-test & >= "0.8.1"}
   "result"
-  "fmt"
 ]
 
 synopsis: "Library for parsing GraphQL queries"
--- a/graphql_parser/src/ast.ml
+++ b/graphql_parser/src/ast.ml
@@ -96,8 +96,10 @@ type document =
   definition list
 
 module Pp = struct
-  let comma = Fmt.(const string ",")
-  let colon = Fmt.(const string ":")
+  let fprintf = Format.fprintf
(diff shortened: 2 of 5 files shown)


andreas:
Looks good, thanks! 👍 I'm happy to merge this when 0.9.0 is merged in OPAM -- just in case there are any last minute changes required before then.

hannesm:
what is the advantage hereof? I honestly find the Fmt combinators much easier to read than Format (also, the diff +100 -84 shows that there is actually more code involved).

keleshev:
@hannesm the advantage is dubious, honestly. Readability is subjective. I find the `Format` style more readable, but whatever the maintainers find most suited to their needs.

The +100 -84 diff is mostly due to more liberal line breaks:

```ocaml
(* Before *)
Fmt.fmt "%a %s%a%a %a" fmt pp_optype op.optype name variables op.variable_definitions directives op.directives selection_set op.selection_set

(* After *)
fprintf ppf "%s %s%a%a %a"
  (optype_to_string op.optype)
  name
  pp_variables op.variable_definitions
  pp_directives op.directives
  pp_selection_set op.selection_set
```

I find that `Format` is adequate, except for hideous names, which you can fix by doing "import as":

```ocaml
let list = Format.pp_print_list
```

andreas:
I don't mind much either way. I hope this can be motivating for further contributions though, so I'm happy to indulge 😄

PR #144, opened 2019-03-03
Replace str with re

andreas:
`Str` is not available on MirageOS freestanding, so replacing with `Re`.

See https://github.com/mirage/irmin/issues/632

### DIFF
--- a/graphql_parser/src/ast.ml
+++ b/graphql_parser/src/ast.ml
@@ -99,16 +99,16 @@ module Pp = struct
   let comma = Fmt.(const string ",")
   let colon = Fmt.(const string ":")
 
-  let quote_string str =
-    let open Str in
-    str
-    |> global_replace (regexp "\\") "\\\\\\\\"
-    |> global_replace (regexp "\"") "\\\""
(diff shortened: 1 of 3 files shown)


hannesm:
Thanks @andreas, for further reading:
- the initial blog post about `re` https://mirage.io/blog/ocaml-regexp
- removing the `Str` module from MirageOS https://github.com/mirage/mirage/issues/663

graphql_parser/src/ast.ml
```diff
@@ -99,16 +99,16 @@ module Pp = struct
...
-    |> global_replace (regexp "\b") "\\b"
-    |> global_replace (regexp "\012") "\\f"
-    |> global_replace (regexp "\n") "\\n"
-    |> global_replace (regexp "\r") "\\r"
-    |> global_replace (regexp "\t") "\\t"
+  let quote_string s =
+    let open Re in
+    s
+    |> replace_string (compile (char '\\'))   ~by:"\\\\"
+    |> replace_string (compile (str "\""))    ~by:"\\\""
```
hannesm: IIUC, you could as well use `char '"'` here.

PR #145, opened 2019-03-21
Fix Yojson compilation warnings

mpcjanssen:


anmonteiro:
This is an unfortunate change in Yojson because it’s not backwards compatible. I think until the `json` type really goes away we should support older versions of yojson. 

For example, I think there isn’t yet a pox_deriving_yojson release that supports the new Yojson types.

mpcjanssen:
In that case it would make sense to allow warnings. As it it, building the example fails with deprecation errors.

Adding the following snippet to the dune file at least makes `dune exec examples/server.exe` work:

```
 (env
    (dev
      (flags (:standard -warn-error -A))))
```

anmonteiro:
@mpcjanssen the point of the dev profile is to have warnings enabled. You can add `--profile=release` when building, or simply disable warning 3 `(flags (:standard -w -3))`

mpcjanssen:
My point is that the current example instructions lead to a failing build. Any fix would be ok.

anmonteiro:
Disabling warning 3 sounds sensible to me.

mpcjanssen:
Disabling the specific warnings makes more sense than this PR.

PR #146, opened 2019-03-21
Remove Str from cohttp-graphql

zshipko:
Found this while experimenting with `irmin-mirage`

`graphql-cohttp` will need an update prior to the Irmin 2.0 release. I will add a note about this in the release issue (mirage/irmin#658).


andreas:
Thanks!

PR #149, opened 2019-03-27
Fix path matching in graphql-cohttp

anmonteiro:


andreas:
Thanks for pointing this out!! 🙏

I noticed visiting `/graphql/` does not work with this solution, so opted for a slightly different one (#150).

PR #150, opened 2019-03-28
Fix routing in graphql-cohttp

andreas:
#146 inadvertently introduced a bug due to the different handling of empty parts when splitting strings:

```ocaml
(* Old implementation *)
Str.(split (regexp "/") "/graphql"
- : ["graphql"] : string list

(* New implementation *)
Astring.String.cuts ~sep:"/" "/graphql"
- : [""; "graphql"]

Astring.String.cuts ~sep:"/" "/graphql/"
- : [""; "graphql"; ""]
```

By adding `~empty:false`, the old behavior can be restored:

```ocaml
Astring.String.cuts ~empty:false ~sep:"/" "/graphql"
- : ["graphql"]

Astring.String.cuts ~empty:false ~sep:"/" "/graphql/"
- : ["graphql"]
```

This is preferable over changing the pattern matching (#149) due to handling of the trailing slashes.


phated:
@andreas and @anmonteiro thanks!  I was smashing my head against this for so long before I reached out. Glad this is getting a fix 😄

andreas:
@phated Sorry about that 😢

PR #154, opened 2019-04-02
ocaml 4.03.0 is required, which defines type result in Pervasives -- no need to open Result!

hannesm:


hannesm:
see https://github.com/ocaml/opam-repository/pull/13803 https://github.com/ocaml/opam-repository/pull/13808

andreas:
Thanks!

PR #156, opened 2019-04-14
graphql-cohttp: Improve parameter handling

andreas:
This PR updates the parameter handling (extracting the query, variables and operation name from the HTTP request and body) in an attempt to match [the description on graphql.org](https://graphql.org/learn/serving-over-http/) and [`express-graphql`](https://github.com/graphql/express-graphql).

In essence, the parameters are extracted from both the query string and the HTTP body, and the two are merged with the former taking precedence.

Further, GraphiQL is not served unless the request explicitly requests HTML to allow get requests.

Remaining tasks:

- [x] Add more test cases
- [x] More testing in GraphiQL

### DIFF
--- a/graphql-cohttp/test/dune
+++ b/graphql-cohttp/test/dune
@@ -1,0 +1,4 @@
+(tests
+  (package graphql-cohttp)
+  (names test)
+  (libraries cohttp-lwt-unix digestif.ocaml graphql-lwt graphql-cohttp alcotest))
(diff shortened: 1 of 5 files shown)


graphql-cohttp/test/dune
```diff
@@ -0,0 +1,4 @@
+(tests
+  (package graphql-cohttp)
+  (names test)
+  (libraries cohttp-lwt-unix digestif.ocaml graphql-lwt graphql-cohttp alcotest))
```
anmonteiro: I think this makes it incompatible with Mirage?

anmonteiro: ooops, meant to comment on the `.opam` file diff...

andreas: These are just test dependencies, so I believe it should be fine?

PR #158, opened 2019-05-02
Use default variables in operations

anmonteiro:
This diff fixes the bug where the following valid query was returning an error saying the variable `first` was missing:

```graphql
query example(
  $first: String! = 1
) {
  greeter(first:$first)
}
```


andreas:
Thanks!

PR #168, opened 2019-08-28
Merge fields defined in fragments

### LINKED ISSUE #167: Fields defined in fragments don't merge properly
If fields are defined in a fragment, they won't be merged properly with the rest of the fields.

Example:

```
query Films {
  allFilms {
    films {
      title
    }
  }
  ...F
}

fragment F on Root {
  allFilms {
    films {
      episodeID
    }
  }
}
```

Should return (see [swapi example](https://graphql.org/swapi-graphql/?query=query%20Films%20%7B%0A%20%20allFilms%20%7B%0A%20%20%20%20films%20%7B%0A%20%20%20%20%20%20title%0A%20%20%20%20%7D%0A%20%20%7D%0A%20%20...F%0A%7D%0A%0Afragment%20F%20on%20Root%20%7B%0A%20%20allFilms%20%7B%0A%20%20%20%20films%20%7B%0A%20%20%20%20%20%20episodeID%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D))
```
{
  "data": {
    "allFilms": {
      "films": [
        {
          "title": "A New Hope",
          "episodeID": 4
        },
      ]
    }
  }
}
```

But in ocaml-graphql-server, it will drop the episodeID
```
{
  "data": {
    "allFilms": {
      "films": [
        {
          "title": "A New Hope"
        },
      ]
    }
  }
}
```

Created a branch with a failing test here: https://github.com/andreas/ocaml-graphql-server/compare/master...dwwoelfel:fragment-handling

dwwoelfel:
Fixes #167.

Merges selection sets in `collect_fields`.

Uses a StringMap to facilitate the merging, then walks the list of fields to ensure we haven't changed the ordering.

### DIFF
--- a/graphql/src/graphql_schema.ml
+++ b/graphql/src/graphql_schema.ml
@@ -1297,6 +1297,35 @@ end
       | `Skip -> Ok false
       | `Include -> should_include_field ctx rest
 
+  let alias_or_name : Graphql_parser.field -> string = fun field ->
+    match field.alias with
+    | Some alias -> alias
+    | None -> field.name
+
+  let merge_selections (fields: Graphql_parser.field list) =
+    let m = List.fold_left (fun m (field : Graphql_parser.field) ->
+      StringMap.update
+        (alias_or_name field)
+        (function
+          | None -> Some field.selection_set
+          | Some x -> (Some (List.append x field.selection_set))
+        )
+        m
+      )
+      StringMap.empty
+      fields
+    in
+    let (_, res) = List.fold_right (fun (field : Graphql_parser.field) (m, res) ->
+      let name = (alias_or_name field) in
+      match (StringMap.find name m) with
+      | None -> (m, res)
+      | Some selection_set -> ((StringMap.remove name m), { field with selection_set = selection_set }::res)
+      )
+      fields
+      (m, [])
+    in
+    res
+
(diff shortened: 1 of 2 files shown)


andreas:
Thanks for looking into this! I tried coming up with something simpler, though asymptotically less efficient:

```ocaml
  let rec merge_selections ?(memo=[]) = function
    | [] -> List.rev memo
    | field::fields ->
        let name = alias_or_name field in
        let matching, rest = List.partition (fun field' -> name = alias_or_name field') fields in
        let selection_sets = List.map (fun (field : Graphql_parser.field) -> field.selection_set) (field::matching) in
        let selection_set  = List.concat selection_sets in
        merge_selections ~memo:({field with selection_set}::memo) rest
```

With both your implementation, and the implementation above, the merged selections are folded into a single field. This is generally not a problem, except this field is exposed to the resolver via [`resolve_info`](https://github.com/andreas/ocaml-graphql-server/blob/e6ad431c0cc1119ecc07a2d6dfcfa7ed19f48d20/graphql/src/graphql_intf.ml#L125). We probably need to change this to be a of list fields. This appears to be consistent with [`graphql-js`](https://github.com/graphql/graphql-js/blob/6f341a3ae0d6b3cc4d9136a58fd9272c615c2dba/tstypes/execution/execute.d.ts#L142).

andreas:
Thanks for your contribution!

PR #170, opened 2019-09-15
Fix for fmt >= 0.8.7

hongchangwu:
This will be unnecessary if #140 is merged. But until then, the constraint on `fmt` is quite annoying and can be easily fixed by adding explicit type annotations.


andreas:
Thanks!

hongchangwu:
Thanks! When you get the chance can you publish a new version to opam?

andreas:
> Thanks! When you get the chance can you publish a new version to opam?

You can follow the opam PR here: https://github.com/ocaml/opam-repository/pull/14873

PR #172, opened 2019-09-20
Add logging to the example

naartjie:
This PR adds logging to `./examples/server.exe` and outputs a where the server URL:

```
listening on http://localhost:8080/graphql
```

Hoping this makes it more intuitive for onboarding newcomers.

### DIFF
--- a/examples/server.ml
+++ b/examples/server.ml
@@ -109,18 +109,17 @@ let schema = Schema.(schema [
 module Graphql_cohttp_lwt = Graphql_cohttp.Make (Graphql_lwt.Schema) (Cohttp_lwt_unix.IO) (Cohttp_lwt.Body)
 
 let () =
+  let open Printf in
   let on_exn = function
     | Unix.Unix_error (error, func, arg) ->
-      Logs.warn (fun m ->
-        m  "Client connection error %s: %s(%S)"
+      printf "Client connection error %s: %s(%S)"
           (Unix.error_message error) func arg
-      )
-    | exn -> Logs.err (fun m -> m "Unhandled exception: %a" Fmt.exn exn)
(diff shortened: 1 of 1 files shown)


andreas:
This is a nice improvement for newcomers. I'm wondering whether simply printing the suggested line to stdout would suffice rather than introduce `logs` as a dependency. Did you have something particular in mind when choosing to use `logs`?

naartjie:
Oh, I used `logs` mainly because it was already in [`examples/server.ml`](https://github.com/andreas/ocaml-graphql-server/blob/8fcae15e53f600f428c88d0c77b0fa3d5baa85a9/examples/server.ml#L114-L118) but I don't think they weren't turned on anyways - I don't think anything would get printed on an error the way it was.

I've replaced both the uses of logs: mine, and what was already there.

examples/server.ml
```diff
@@ -109,16 +109,17 @@ let schema = Schema.(schema [
...
-      )
-    | exn -> Logs.err (fun m -> m "Unhandled exception: %a" Fmt.exn exn)
+    | exn -> printf "Unhandled exception: %s\n%!" (Printexc.to_string exn)
   in
   let callback = Graphql_cohttp_lwt.make_callback (fun _req -> ()) schema in
   let server = Cohttp_lwt_unix.Server.make_response_action ~callback () in
-  let mode = `TCP (`Port 8080) in
+  let port = 8080 in
+  let mode = `TCP (`Port port) in
+  printf "listening on http://localhost:%d/graphql\n%!" port ;
```
andreas: Nit (... if only we had automated code formatting 😉)

```suggestion
  printf "listening on http://localhost:%d/graphql\n%!" port;
```

andreas:
thanks!

PR #175, opened 2019-09-22
A preview of ocamlformat, using conventional profile (don't merge)

### LINKED ISSUE #173: ocamlformat
Would you accept a PR to add `ocamlformat` to the project? If so, any preference for profile?

I was thinking of just adding the project files, and @andreas could run `dune build @fmt --auto-promote`. I wouldn't want to be the main committer.

Anyways, let me know if you'd be keen, if so, I can do a couple of disposable PR's with the different profiles, just to preview what the code would end up like with each of them.

naartjie:
fixes #173 

This is a preview of `ocamlformat` using the `conventional` profile.

### DIFF
--- a/graphql/test/schema_test.ml
+++ b/graphql/test/schema_test.ml
@@ -1,340 +1,401 @@
 let test_query = Test_common.test_query Test_schema.schema ()
 
-let suite = [
-  ("query", `Quick, fun () ->
-    let query = "{ users { id } }" in
-    test_query query (`Assoc [
-      "data", `Assoc [
-        "users", `List [
-          `Assoc [
-            "id", `Int 1
-          ];
-          `Assoc [
-             "id", `Int 2
-          ]
-        ]
-      ]
-    ])
-  );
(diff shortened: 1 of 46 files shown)


graphql/test/schema_test.ml
```diff
@@ -1,340 +1,401 @@
...
+            ]) );
+    ( "introspection query should be accepted",
+      `Quick,
+      fun () ->
+        let query =
+          "\n\
+          \      query IntrospectionQuery {\n\
+          \        __schema {\n\
+          \          queryType { name }\n\
+          \          mutationType { name }\n\
```
naartjie: I'm not sure if there is something funky happening on these lines @andreas. I don't know this syntax (don't understand it). The tests still pass though.

andreas: It needs to be rewritten to use `{| ... |}` rather than `" ... "` to avoid the ugly line breaks.

naartjie:
Closing this one, as it was only a preview, and the real PR is #177

PR #177, opened 2019-10-11
Add ocamlformat

naartjie:
This PR adds `ocamlformat`, using the `conventional` format.

It only brings the minimum changes necessary to format the project:

```sh
$ dune build @fmt --auto-promote
```

After that it is expected that @andreas would commit those changes as the main author. After that we would enforce the formatting rules on travisCI (as per [@anmonteiro's suggestion](https://github.com/andreas/ocaml-graphql-server/issues/173#issuecomment-536330693)) to ensure that any new code merged follows the formatting rules.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,1 @@
+profile = conventional
--- a/dune-project
+++ b/dune-project
@@ -1,2 +1,3 @@
 (lang dune 1.1)
 (using menhir 2.0)
+(using fmt 1.2)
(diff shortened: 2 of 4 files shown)


andreas:
@naartjie, thanks a lot! This has been super helpful. I probably wouldn't have gotten around to doing it at this point otherwise, so really appreciate it 🙏

naartjie:
Awesome @andreas. I'm super happy that I could help.

Should I take a look at setting up the Travis CI hook next - i.e. fail if the code is not formatted?

PR #178, opened 2019-10-12
Fix 2 subscriptions bugs

anmonteiro:
1. Instrospection on `__typename`
2. No error was returned if requesting a subscription field that didn't
exist


andreas:
Thanks! It would be nice if `resolve_fields` could handle subscription objects, too.

anmonteiro:
Agreed -- I can try to unify these two functions but currently the types don't match so that will probably require some more work.

PR #183, opened 2019-12-18
Add introspection fields during schema creation

dwwoelfel:
This solves a problem we saw at OneGraph when upgrading to the latest version of ocaml-graphql-server.

We saw our query times spike by ~60ms on every query after upgrading. We tracked down the cause to this commit https://github.com/andreas/ocaml-graphql-server/commit/b187eaa15046e8fc01b111a8ac4791d6c92a0e3b, which calls `types_of_schema` on every call to `execute`.

With this change, `types_of_schema` is calculated only once when you construct the schema.

### DIFF
--- a/graphql/src/graphql_schema.ml
+++ b/graphql/src/graphql_schema.ml
@@ -514,30 +514,6 @@ module Make (Io : IO) (Field_error : Field_error) = struct
     subscription : 'ctx subscription_obj option;
   }
 
-  let schema ?(mutation_name = "mutation") ?mutations
-      ?(subscription_name = "subscription") ?subscriptions
-      ?(query_name = "query") fields =
-    {
-      query =
-        {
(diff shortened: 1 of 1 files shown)


graphql/src/graphql_schema.ml
```diff
@@ -1966,7 +1940,43 @@ module Make (Io : IO) (Field_error : Field_error) = struct
...
+                abstracts = ref [];
+                fields = lazy fields;
+              });
+        subscription =
+          Option.map subscriptions ~f:(fun fields ->
+              { name = subscription_name; doc = None; fields });
+      }
+    in
+    let types = Introspection.types_of_schema schema in
+    Introspection.add_built_in_fields schema types
```
andreas: How about leaving `add_built_in_fields` as is and simply call it from here?

Issue #31, opened 2017-02-26
Introspection query should contain argument default value


andreas:
Currently the `defaultValue`-field of `__InputValue`-type is always `null` in introspection query results. It should be populated as [per the spec](https://facebook.github.io/graphql/#sec-The-__InputValue-Type).

dwwoelfel:
I've been looking at this, and I think the best approach would be to change the type of `~default` in `arg'` from `'a` to `Graphql_parser.const_value`. From there, you can call `string_of_const_value` to get the `defaultValue` field in the introspection query and `eval_arg` to get the same value of type `'a` that is currently passed to `~default`.

Example

Before
```ocaml
arg' "someArg" ~typ:string ~default:"defaultString"
```

After
```ocaml
arg' "someArg" ~typ:string ~default:(`String "defaultString")
```

I looked at deriving a `Graphql_parser.const_value` from the `~default` arg, but the `~coerce` function for an input object loses information, so I don't think it would be possible to derive the `const_value`.

Example:

```ocaml
let argObj =
  Arg.obj "InputObject"
     ~fields:Arg.[(arg "argOne" ~typ:int); (arg "argTwo" ~typ:int)] 
     ~coerce:(fun arg_one arg_two -> arg_one + arg_two) in
arg' "inputArg" ~typ:argObj ~default:3
```

That `defaultValue` in that case could be either `{a: 1 b: 2}` or `{a: 3 b: 0}`.

With the default as const_value approach, this would be defined as

```ocaml
arg' "inputArg" ~typ:argObj ~default:(`Assoc [("a", `Int 1); ("b", `Int 2)])
```

One downside of this approach is that you can't ensure that `~default` is correct at compile-time. I think it should be possible at schema construction time or at introspection-query time at the very least.

PR #194, opened 2020-08-17
Show defaultValue in introspection query

dwwoelfel:
Per https://github.com/andreas/ocaml-graphql-server/issues/31#issuecomment-674311844, changes the type of the `~default` argument in `arg'` to `Graphql_parser.const_value` to make it possible to populate the `defaultValue` field in the introspection query.

### DIFF
--- a/examples/server.ml
+++ b/examples/server.ml
@@ -86,7 +86,7 @@ let schema =
                           ~coerce:(fun greeting name -> (greeting, name))
                           ~fields:
                             [
-                              arg' "greeting" ~typ:string ~default:"hello";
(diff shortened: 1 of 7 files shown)


andreas:
Thanks for writing up #31 and implementing this change 🙏 

From a conceptual point of view, I would prefer an approach where defining an input scalar (i.e. calling `Arg.scalar`) would require a providing function `'a -> Graphql_parser.const_value`. However, as you note in #31, there are some difficulties with this for input object types that I haven't worked out yet. What you propose seems quite pragmatic until that's sorted out.

PR #198, opened 2021-01-29
Support block strings in the parser

dwwoelfel:
Adds parsing support for GraphQL's block string syntax.

With this change, the message arg in the following query will be parsed as `"Hello,\n  World!\n\nYours,\n  GraphQL."`, per the [spec](https://spec.graphql.org/draft/#sec-String-Value.Block-Strings).

```graphql
mutation {
  sendEmail(message: """
    Hello,
      World!

    Yours,
      GraphQL.
  """)
}
```

Test cases are lifted from the [block string tests in graphql-js](https://github.com/graphql/graphql-js/blob/main/src/language/__tests__/blockString-test.js).

### DIFF
--- a/graphql_parser/src/block_string.ml
+++ b/graphql_parser/src/block_string.ml
@@ -1,0 +1,66 @@
+let get_block_string_indentation (s : string) : int =
+  let is_first_line = ref true in
+  let is_empty_line = ref true in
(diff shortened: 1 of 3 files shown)


andreas:
Thanks @dwwoelfel! 🏆

PR #200, opened 2021-02-24
Fix skip/include directives on fragment spreads

dwwoelfel:
Quick patch for skip/include directives on fragment spreads. 

The directives weren't being applied to the fragment spread because it was looking at the directives that were defined on the fragment definition (where skip and include aren't allowed).

In the example below, it was looking at (the made-up) `fragmentDefDirective` instead of the `skip` directive.

```
query TestQuery {
  test {
    ...TestFragment @skip(if: true)    
  }
}

fragment TestFragment on Test @fragmentDefDirective {
  id
}
```

Includes a test.


anmonteiro (approved):
Thanks for fixing this!

andreas:
Thanks @dwwoelfel!

PR #201, opened 2021-06-23
Add GH Actions CI.

tmcgilchrist:

PR #202, opened 2021-08-16
Add GH Actions CI enabling various OCamls.

tmcgilchrist:
No windows builds because Async doesn't support Windows so the async bindings won't build there.
Removes travis since that isn't working anymore.

Previous PR that I closed with some comments https://github.com/andreas/ocaml-graphql-server/pull/201


andreas:
Thanks, @tmcgilchrist! I'll take a look later this week so we can get it merged 🙏

tmcgilchrist:
You can see a CI from the first commit here https://github.com/tmcgilchrist/ocaml-graphql-server/runs/3336582952
The second commit builds everything (including examples with a code fix included) and pulls in a `js_of_ocaml` dependency, which I'm not sure where that comes from.

andreas:
It seems like the build for d83e9c59c2799c1bb7305db2b1a622ed030fe385 passes, but the build for 0a03d35833a7df706ed1897ac595ffda5df9a23d does not despite your fix?

As such, I would prefer merging the former over the latter, even if it doesn't include the examples. The Travis setup did not include the examples either, but it would be great to do at some point 🙂

tmcgilchrist:
I've just included d83e9c5 in this PR.
I am still not sure where the `js_of_ocaml` dependency is coming from, but the code fix to the example is valid. 
I'll submit that as it's own PR.

andreas:
Thanks for contributing this and sorry for the delay in merging! [The run](https://github.com/andreas/ocaml-graphql-server/actions/runs/1270954280) passed successfully 🥳

Issue #196, opened 2020-09-18
The server will reject the request from okhttp and dart-lang http


crackerli:
hi, andreas:

Some of the http lib like okhttp, dart-lang http will append "charset=utf-8" to header "Content-Type" automatically, it will cause the rejection of ocaml-graphql-server, but pity hard to remove it from "Content-Type". Now I use apollo-android and graphql-flutter to develop some app connect to ocaml-graphql-server, I find no workaround to fix it by my side.

Maybe it is easy for server side to change? Sorry for I have on ocaml knowledge.
I just add some test code like below, and it works for me in my app.

![image](https://user-images.githubusercontent.com/10629416/93617842-13ad5e00-fa09-11ea-9648-1e6e78f678c5.png)

andreas:
Hey @crackerli - yeah, this should be fixed server side. It looks like Cohttp has a method to extract the content type without the charset: `Header.get_media_type`.

Thanks for reporting 🙂

crackerli:
Thank you much for your quick reponse, andreas

crackerli:
hi, andreas, when this issue can be fixed?

pm-mck:
I made a PR #204 to address this issue (we just hit it too while making a Terraform provider). Let me know if there's anything else needed to get it merged in.

patricoferris:
#204 was merged, so this should be fixed now thanks @pm-mck

PR #204, opened 2021-12-11
Fixes content-type + encoding issue

pm-mck:
fix: issue where Fmt.prefix is deprecated, replaced with ( ++ ); 

fix: issue andreas/ocaml-graphql-server#196 where content-type containing an encoding would fail to process

### DIFF
--- a/graphql_parser/src/ast.ml
+++ b/graphql_parser/src/ast.ml
@@ -138,7 +138,7 @@ module Pp = struct
 
   and selection_set fmt =
     omit_empty_list
-      Fmt.(braces (hvbox ~indent:2 (prefix cut (list pp_selection))))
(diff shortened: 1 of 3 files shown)


graphql_parser/src/ast.ml
```diff
@@ -138,7 +138,7 @@ module Pp = struct
 
   and selection_set fmt =
     omit_empty_list
-      Fmt.(braces (hvbox ~indent:2 (prefix cut (list pp_selection))))
+      Fmt.(braces (hvbox ~indent:2 (cut ++ (list pp_selection))))
```
pm-mck, line 141: This was necessary to change due to `Fmt` deprecating `prefix`.

samoht:
Thanks!

PR #208, opened 2022-05-31
Use Yojson `t` types instead of deprecated `json`

Leonidas-from-XIV:
This fixes compatibility for the upcoming Yojson 2 while still working with Yojson >= 1.6.


andreas:
Thanks!

PR #209, opened 2022-06-06
Raise minimum `rresult` version

Leonidas-from-XIV:
The code needs at least version 0.2.0 of `rresult` for `>>=` and it needs at least 0.3.0 for `rresult` to use the `result` compatibility package to define a `result` type that is compatible with the stdlib `result` type.

Discovered via the minimum-version checks of https://github.com/ocaml/opam-repository/pull/21475


andreas:
Thanks again! 🙂

PR #220, opened 2025-12-03
set Content-Type for graphiql html

zshipko:


zshipko:
Seems like https://github.com/andreas/ocaml-graphql-server/pull/219 would fix the CI failures

